PR #28, opened 2015-12-11
Added translated text selection for Windows Phone IE

drakosvlad:


sushain97:
Please don't use any inline JS. Instead, use listeners (examples in `/assets/js/`).

sushain97:
Closing PR since there seems to have been no explanation as to why this works or improvements of code style.

PR #30, opened 2015-12-16
Added UI for suggesting a word

svineet:


sushain97:
Your constant height is causing [problems](https://skc.name/u?nIvN.d). Use some `overflow` CSS. Also need a `title` on each of the words that replacements can be suggested, i.e. "Improve Apertium's translation". Also, this needs to be localizable along with the strings in the modal. Feel free to send multiple commits instead of changing the same one.

image

sushain97:
Also, try to see if you can position the modal such that the source text is still visible while maintaining responsive design. If this isn't possible, then it's fine.

PR #32, opened 2016-01-01
UI changes and connecting Suggest UI to APY endpoint

svineet:
This is still a work in progress (almost done, a few refinements required). 

Uploading so that others can test for bugs as well. 

[task](https://codein.withgoogle.com/dashboard/task-instances/5765903643836416/?sp-page=1)


sushain97:
There are ways to get the text without using hashes. First suggestion that comes to mind is selecting the text then getting the indexes of that into the whole text. This may be helpful: http://stackoverflow.com/questions/2828713/javascript-i-can-get-the-text-of-a-selection-now-how-do-i-get-the-text-outside?rq=1 for another approach.

PR #34, opened 2016-01-05
Fix deprecation warnings for webkitURL and XMLHttpRequest event properties

svineet:
Deprecation warning for `ev.totalSize` still persists, even though the warning for `ev.position` is gone, so I removed totalSize completely


sushain97:
Did you verify that removing it still works cross-browser? Perhaps use the
|| technique here as well to escape deprecation warnings.

svineet:
Yes, tried it on Firefox as well. The problem is that doing || still throws the deprecation warning in Chrome. I don't know why.

svineet:
OK I found another warning I hadn't seen before about persistence, fixed it.

PR #35, opened 2016-01-05
Suggestions

sushain97:


unhammer:
did you try this on turkic? @jonorthwash e.g. with

     git fetch origin suggestions
     git checkout -b suggestions origin/suggestions

svineet:
OK so to run the code, a ReCAPTCHA site key and the secret that comes along with that, credentials for a wiki account and the wiki page the data should be recorded in are required.

Run apy with command:

    ./servlet.py /usr/local/share/apertium/ --wiki-username=WikiUsername --wiki-password=WikiPassword -rs=YourRecaptchaSecret
Since I'm bringing this up, it would be nice if APY could accept a configuration file, would make these commands shorter.

And copy the new `config.example.conf` file to `config.conf` and edit the following properties:

    [SUGGESTIONS]
    RECAPTCHA_SITE_KEY = YourSiteKey
    CONTEXT_WRAP = 5

`CONTEXT_WRAP` is the number of words that the context takes from each side of the `targetWord`, so you can get longer contexts or shorter contexts as required.

The reCAPTCHA site key and secret can be obtained by adding your own entry here: https://www.google.com/recaptcha/admin#list

reCAPTCHA already adds `localhost` as an allowed domain for requests so no need to add your (or your server's) IP address for testing, but for production it is required.

jonorthwash:
Is there any reason some of the information from the other switches can't be included in the conf file?  Also, do we want some official org registration with ReCAPTCHA, or should I just make one myself for turkic.apertium.org (as its de facto admin)?

unhammer:
@jonorthwash see https://github.com/goavki/apertium-apy/issues/5 about conf; about registration I'd say make one yourself. An official org registration would be warranted if it could be seen as somehow representing Apertium, but this is just about getting an API key, right?

svineet:
An individual account should be fine, only a site key and the secret is needed. You can always switch to another api key if need be, it won't be a problem.

jonorthwash:
What needs to be done here still?  Can this be merged?

svineet:
Let's not merge this now, merge apy first (I will have to resolve merge conflicts for that PR). The UI could use some refinement on this, but minor UI changes aside as a whole I think it's ready. Let's list down the things we need to figure out, anything you guys can think about, I'll get to it this weekend.

jonorthwash:
My main goal is to get your code merged in and the PR requests closed, so whatever the minimum is to do that is fine :)

PR #39, opened 2016-01-12
Add 'q' parameter to URL

### LINKED ISSUE #14: q=input+text
It'd be cool if the contents of the q parameter were put into the input text box. 

I'm not 100 % sure if we should do like google and update it in the URL (gets a bit ugly for large texts, maybe only for small ones?)

svineet:
Fix #14

My internet has been acting up lately, sorry if I forgot something.

[Task](https://codein.withgoogle.com/dashboard/task-instances/6463955165773824/?sp-page=1)

### DIFF
--- a/assets/js/localization.js
+++ b/assets/js/localization.js
@@ -60,14 +60,14 @@ $(document).ready(function () {
         var pathname = window.location.pathname;
 
         if(window.history.replaceState && !pathname.endsWith('/index.debug.html')) {
-            var urlParams = [], urlParamNames = ['dir', 'choice'];
+            var urlParams = [], urlParamNames = ['dir', 'choice', 'q'];
             $.each(urlParamNames, function () {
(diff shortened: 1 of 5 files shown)


assets/js/localization.js
```diff
@@ -60,14 +60,14 @@ $(document).ready(function () {
         var pathname = window.location.pathname;
 
         if(window.history.replaceState && !pathname.endsWith('/index.debug.html')) {
-            var urlParams = [], urlParamNames = ['dir', 'choice'];
+            var urlParams = [], urlParamNames = ['dir', 'choice', 'q1', 'q2', 'q3'];
```
sushain97: Use `q` (translation), `qA` (analysis), and `qG`.

unhammer: Is there any reason we can't use q for all of them? You're only in one mode (trans/gen/ana) at any one time.

PR #53, opened 2016-03-06
Kira's GSoC Coding challenges: Improve translation button added

Kira-D:
This is a pull request with GSoC Coding challenges task (as agreed with Jonathan)


jonorthwash:
We should have this button so it can be enabled or disabled by the site maintainer in the [config](https://github.com/goavki/apertium-html-tools/blob/master/config.conf.example) file, like for translation, analysis, sandbox, etc.

Kira-D:
"Improve_Translation" label added to eng.json + 1 missing label for Russian

jonorthwash:
I think this PR can be closed and merged in to the main branch at this point.  @unhammer, what do you think?

PR #73, opened 2016-08-28
Detection

Kira-D:


assets/js/translator.js
```diff
@@ -614,7 +618,9 @@ function translateDoc() {
     }
 }
 
-function detectLanguage() {
+function identify(detect=false) {
```
sushain97: Please don't use default parameters: https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Functions/default_parameters#Browser_compatibility. Hound is complaining because ES6 isn't enabled. `detect` falsey by default if you don't pass in a value so this should be a very simple fix.

assets/js/translator.js
```diff
@@ -120,10 +125,15 @@ if(modeEnabled('translation')) {
...
-            detectLanguage();
-            translateText();
+            if ($('#originalText').val() != '') {
+                refreshLangList(true);
+                $('.srcLang').removeClass('active');
+                $(this).addClass('active');
+                detectLanguage();
+            }
+            else {
+                console.error('Empty textarea');
```
sushain97: Just remove this branch.

assets/js/translator.js
```diff
@@ -4,6 +4,9 @@ var curSrcLang, curDstLang;
 var recentSrcLangs = [], recentDstLangs = [];
 var droppedFile;
 var textTranslateRequest;
+var nonDetectButton;
+//var identifiedCode;
```
sushain97: Remove.

assets/js/translator.js
```diff
@@ -4,6 +4,9 @@ var curSrcLang, curDstLang;
...
+globalImplementedLangs = [];//this is global specially for identify()
```
sushain97: Why not use `srcLangs`...?

assets/js/translator.js
```diff
@@ -630,41 +636,84 @@ function detectLanguage() {
...
+                        handleNewCurrentLang(curSrcLang = identifiedCode, recentSrcLangs, 'srcLang');
+                        autoSelectDstLang();
+                    });
                 }
             }
-            recentSrcLangs = recentSrcLangs.concat(oldSrcLangs);
-            if(recentSrcLangs.length > 3) {
-                recentSrcLangs = recentSrcLangs.slice(0, 3);
+            else {
+                $('#identify').html(dynamicLocalizations['Sorry_Support'] + langNameTarget + ', ' + '<a data-toggle="modal" data-target="#contactModal" data-keyboard="true" style="cursor: pointer">' + dynamicLocalizations['ContactUs'] + '</a>');
```
sushain97: Would this code ever execute? o.o

sushain97:
I have no idea what this PR actually does and it's super stale :(

PR #78, opened 2016-09-23
Context size fixed

Kira-D:


sushain97:
@shardulc I think this PR needs to be de-duped and merged in as well for suggestions.

PR #93, opened 2017-01-09
Fix minor issues with the chaining functionality

shardulc:


assets/js/translator.js
```diff
@@ -730,7 +731,7 @@ function detectLanguage() {
 }
 
 function translationNotAvailable() {
-    $('#translatedText').text(dynamicLocalizations['Not_Available']);
+    $('#translatedText').val(dynamicLocalizations['Not_Available']);
```
sushain97: Can we actually just change both val and text? I'm not sure of the intricacies behind this and it won't hurt.

assets/js/translator.js
```diff
@@ -20,7 +20,8 @@ if(modeEnabled('translation')) {
 
         function getDstLangs(srcLang) {
             var targets = [];
-            var targetsSeen = {srcLang: true};
+            var targetsSeen = {};
+            targetsSeen[srcLang] = true;
```
sushain97: Why did this become two lines? It looks the same??

assets/js/translator.js
```diff
@@ -572,7 +573,7 @@ function inputFile() {
 }
 
 function translateDoc() {
-    var validPair = pairs[curSrcLang] && pairs[curSrcLang].indexOf(curDstLang) !== -1,
+    var validPair = originalPairs[curSrcLang] && originalPairs[curSrcLang].indexOf(curDstLang) !== -1,
```
sushain97: This doesn't change the list, though?

sushain97: This doesn't change the list, though?

assets/js/translator.js
shardulc: Seems to work differently: the old one adds a key called 'srcLang', while the new one actually expands the variable.

assets/js/translator.js
shardulc: We want to preserve the checked/unchecked-ness of the checkbox, right?

assets/js/translator.js
sushain97: OH, right. I've definitely made this same mistake before. Easy bug to miss :)

assets/js/translator.js
sushain97: Yep.

PR #94, opened 2017-01-09
Chained translation

sushain97:


sushain97:
This PR introduces an extra checkbox. An options button with a cog icon should be created that reveals the three options when clicked below it.

sushain97:
@jonorthwash can you test this branch on turkic.apertium.org (`git checkout chained-translation` and update APy)?

Issue #91, opened 2017-01-08
Chained translation path display


sushain97:
It would be cool to have chained translation display the path taken to get from say eng => kaz.

@jonorthwash

Putti:
IMHO, this is not particularly relevant information for the end user.

ftyers:
It's super useful! :D ...  It's useful for determining how reliable the translation is, both in terms of 1) how many steps, 2) which languages it goes through, e.g. spa-cat-ita is more reliable than eus-eng-fin

Putti:
Well, yeah it is useful if you know those things already, so for more advanced user it is quite useful information and thus +1 for this feature from me, too.

sushain97:
I was thinking of the best way to display this... Perhaps this?

![image](https://cloud.githubusercontent.com/assets/4211838/21965813/6de884ae-db35-11e6-8e91-a0db6e9453de.png)

sushain97:
Alternative:

![image](https://cloud.githubusercontent.com/assets/4211838/21965865/74104410-db36-11e6-9165-25aeb4d4e0e0.png)

https://gist.github.com/sushain97/e3a6077da7c21399847a21dd94e3daa3

shardulc:
This was discussed on #apertium intermittently from [19:13:13] on [2017-01-14](http://tinodidriksen.com/pisg/freenode/logs/%23apertium_20170114.log) to [01:18:24] on [2017-01-15](http://tinodidriksen.com/pisg/freenode/logs/%23apertium_20170115.log), and again from [20:14:45] onwards on [2017-01-15](http://tinodidriksen.com/pisg/freenode/logs/%23apertium_20170115.log).

jonorthwash:
What if we have the ‘?’ icon next to the translate button instead of on it?

PR #104, opened 2017-02-13
Initial code for a visual chained translation path selector

shardulc:
Add a button which opens a modal with a graph of the possible paths from the selected source to the selected destination. The button is visible only when the multi-step checkbox is checked. The modal has a rudimentary 'Valid Paths' section with links to select the path: this works, but I'm not good at UI so I'm requesting help. Many of the strings/UI might require change anyway. (Uses d3.) (Ref #91)


sushain97:
@shardulc thanks for working on this!! I will take a look soon hopefully :)

share-with-me:
This feature is awesome :D . I have myself worked quite a bit on d3.js and I would love to take up this feature. Is there some working demo or some visualisation of this d3 code? 
In the description of this issue, it says it has a section of Valid Paths. How do we know that? Is some analysis taking place in the backend which generates a list of valid paths (some graphs/DFS stuff?) ?

sushain97:
The analysis takes place in APy. I believe it's Dijkstra's.

The feature itself is quite cool. You can checkout this PR and try it out!

shardulc:
@share-with-me As @sushain97 said, the default path selection for chained translation is done in APy (specifically [here](https://github.com/goavki/apertium-apy/blob/master/servlet.py#L144)) which is currently just Dijkstra's with equal weights for all pairs. There is a small chance of the weights being significant in the future if we have a good metric for pair quality. However, all possible paths cannot and should not be sent to clients because that is easily computable from just a list of pairs (and potentially weights), so on the html-tools side, there is BFS ([here](https://github.com/shardulc/apertium-html-tools/blob/6996354c977feac8ddb0f2873a5a2ae153f029af/assets/js/translator.js#L355)).

The 'Valid Paths' referred to are basically just paths which include almost all (actually at least all but one) of the nodes that a user selects and go from the source to the destination. This is also easy to compute.

d3 is really cool! Please do checkout this PR and try it out, we can discuss any suggestions/contributions you have.

shardulc:
[Things discussed](http://tinodidriksen.com/pisg/freenode/logs/%23apertium_20170314.log) in `#apertium` on 2017-03-14 at 03:01:11 for around an hour:

- [ ] Modal should have default path already chosen
- [ ] The button should say "refine current path" or something similar, and should have a hover icon
- [ ] 'Valid Paths' should be more prominently displayed, with placeholder text advising users if there are no valid paths
- [ ] Only the relevant arrowheads should be displayed instead of all available ones

shardulc:
I tried to add everything in the checklist from the previous comment, but I just had two questions:
1. What should the 'hover icon' look like?
2. Instead of placeholder text when there is no path selected, the modal starts out with the Valid Paths section indicating the default path, and if the user selects anything, the default path goes away.

sushain97:
^ looks like there are a couple conflicts :(

Issue #79, opened 2016-09-28
414: Request-URI too large (should use POST?)


unhammer:
I tried translating the text of http://www.elespanol.com/ciencia/tecnologia/20160926/158484429_0.html and got "Translation not yet available!", console showed the request was sent as GET and got back a 414.

VictorVelev:
I want to fix it!

share-with-me:
@unhammer , @sushain97 , I have changed the GET method to POST method in translateText() function and it now processess  the bigger amount of data and does not give '414 error'. (SS below) However, I had a few questions:
1. $.jsonp does not work with any method except 'GET'. This is the link to support this argument. [Stackoverflow](http://stackoverflow.com/questions/4508198/how-to-use-type-post-in-jsonp-ajax-call)
2. Now, if we want to use 'POST' to cater to bigger amount of text, we will have to switch to some other method - which is what I have used ($.ajax())
3. I read in the uses of jsonp datatype that it allows transfer of data between domains of different origins. Now, will that be restricted if I use $.ajax() instead of that? Also, does this expose any security threat for the application? Thanks!
I uses some data from the above mentioned website and tested it with GET and POST. Below are screenshots!
![get_query](https://cloud.githubusercontent.com/assets/8894636/23831066/6c85af66-073f-11e7-9714-2aab2780d7ce.png)
![post_query](https://cloud.githubusercontent.com/assets/8894636/23831068/73524c32-073f-11e7-902f-c2a1c2a4d205.png)

share-with-me:
Should we keep an upper limit on the length of input text box and put a feature like overflow-y: scroll? Currently, the size of text box shows no limit and dynamically updates its size with data?

sushain97:
@share-with-me There's no need for a `y-overflow: scroll`. However, I just added #108 which may mitigate the problem you mention.

sushain97:
@share-with-me There's no security threat. However, the limitation you discuss is valid. Great research.

I would suggest that you use POST (btw `$.post` exists) when a 414 would happen (this requires some research) and JSONP otherwise. POST won't work due to cross-domain restrictions for some installations so we don't want to use it by default, only when necessary.

share-with-me:
@sushain97 Sure! I will try this with $.post() too! Thank you for all the information :D . As for task #108, I will get on it. What is bugging me since evening is the right-to-left-ish behaviour of data when cookies are disabled :P

unhammer:
[...]

PR #112, opened 2017-03-13
Changed GET to POST in translateText() function

share-with-me:
Added this functionality currently only in translateText() function. This now ensures that the user can supply large texts in input text box and he wont get 414 error.
Will add this across generator, morphological analysis and analyser mode once I get comfortable with their functioning.

### DIFF
--- a/assets/js/translator.js
+++ b/assets/js/translator.js
@@ -581,28 +581,52 @@ function translateText() {
             }
             request.q = $('#originalText').val(); // eslint-disable-line id-length
             request.markUnknown = $('#markUnknown').prop('checked') ? 'yes' : 'no';
-            textTranslateRequest = $.ajax({
-                url: config.APY_URL + endpoint,
-                beforeSend: ajaxSend,
-                complete: function () {
-                    ajaxComplete();
-                    textTranslateRequest = undefined;
-                },
-                data: request,
-                type: 'POST',
-                contentType: 'application/x-www-form-urlencoded; charset=UTF-8',
-                dataType: 'json',
-                success: function (data) {
-                    if(data.responseStatus === HTTP_OK_CODE) {
-                        $('#translatedText').val(data.responseData.translatedText);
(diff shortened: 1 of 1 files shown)


share-with-me:
The code yields the following functionality:
screen shot 2017-03-13 at 5 01 15 pm

unhammer:
We only want POST requests on long strings – short ones should still go over GET using jsonp, cf https://github.com/goavki/apertium-html-tools/issues/79#issuecomment-286061122

share-with-me:
@unhammer , Ohh Alright! So as this [stackoverflow](http://stackoverflow.com/questions/266322/is-there-a-limit-to-the-length-of-a-get-request) post says, I will keep a threshold for this length around 2000 chars. Anything more than that, I will use $.ajax instead $.post !

share-with-me:
@unhammer  I have added the functionality where larger texts will use $.ajax() and smaller texts will use $.jsonp() as originally used. Screenshots are attached below. The last lines in each cosole image describe the method that is used.
screen shot 2017-03-13 at 5 46 05 pm
screen shot 2017-03-13 at 5 46 36 pm

sushain97:
I would approach implementing this by creating a new function that simply wraps over both `ajax` and `jsonp`, taking a second parameter that indicates the size of the request. Then, this function will send the request. This will make it more DRY especially when you're extending to the other modes.

Also, make sure to take a look at the Travis/Hound violations/build failures on this PR and others.

Good work!

share-with-me:
@sushain97  Yes, surely! I tried to resolve some hound and travis errors. Huff, they are good error pointers :P . However, I will first focus on achieving the functionality and then in a while, solve these errors too. As for the DRY philosophy, on it. I will wrap this function and then call the appropriate method. Thanks for motivating! :D

share-with-me:
Closing this PR as I've created a refactored PR in a separate branch!

PR #113, opened 2017-03-13
Notification if the page takes time to load (greater than a threshold)

share-with-me:
Uses PNotify - the open source tool for generating unobtrusive notices.


sushain97:
@share-with-me I like the aesthetic provided by the PNotify library. However, it's really difficult to justify adding an entire library for this feature. Perhaps you can replicate its functionality (only the necessary bits)? It would be a great learning experience.

share-with-me:
@sushain97  Ohh perfect! I will get on it! Initially I thought by doing the same way but when I saw your suggestion regarding use of the bootstrap plugin, I thought we indeed need some library. I will get on it right away!

share-with-me:
@sushain97 , I have made a custom notice bar and you can check its functionality in the following link. It gets displayed for about 10 seconds (in case user does not close it) enabling the user to read the message completely. Clicking the close button closes the div. Here is the [link](https://drive.google.com/open?id=0BwrSYlOsMCyWMXYycGpabWpWN2s)

sushain97:
Looks very promising! Go ahead and update this PR and I'll have some UI suggestions.

share-with-me:
Closing this pull request here and continuing it in a separate PR ( as the entire interpretation what I had made about this feature was kind of not what was expected for this). Will create a different PR with the functionality for this feature shortly.

PR #117, opened 2017-03-15
Removed Inline Styling

share-with-me:
Removed all the inline styling and created ids or classes (wherever required) and applied styling on the elements. The styling changes can be found in style.css file.

### DIFF
--- a/assets/css/style.css
+++ b/assets/css/style.css
@@ -321,3 +321,167 @@ html[dir=rtl] .pull-right:not(#rtlChecker) {
 html[dir=rtl] .pull-left:not(#rtlChecker) {
     float: right !important;
 }
+
+#wrap-container-1 {
+    position: relative;
+}
+
+.apertiumLogo {
+    margin-bottom: -0.25em;
+}
+
+#globe-div {
+    width: 35%;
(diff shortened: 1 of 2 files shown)


sushain97:
Thanks for working on this! Looks like a good start.

A couple of comments:

1) Since the rest of the classes are `camelCase`, we should stick to that style.
2) We want semantic names for things so tacking on numbers isn't the best way to go.
3) The rules in `styles.css` are currently split by ~mode so we should try to continue to keep them organized (see the comments).
4) Make sure to think about whether we want a class or an id in some cases. If it's a style used multiple times, we want a class. I'm sure you've already thought about this so it's just a quick reminder.

share-with-me:
1. camelCase, yepp will follow that. 
2. Will update the names to hold semantics of elements.
3. Yepp, need to modularise it.
4. This point I took into consideration. In case 2 or more elements requiring same styling, I made a class. In case only a unique element needs it, an id.
Yep, I would update this PR and make it mergeable :D 
Thanks!

sushain97:
Closing for now since it is incomplete. Feel free to re-open later.

PR #119, opened 2017-03-17
Copy button to copy the contents of translatedText

share-with-me:
This button when clicked copies the contents of `translatedText` div. This is useful if the user wants to directly copy it somewhere. A demo can be found [here](https://drive.google.com/open?id=0BwrSYlOsMCyWYlA4eUMxc1NUX2M)


sushain97:
Interesting idea! I do like this however I would also like a flag in the config to enable it showing up. It might also make sense if it were positioned inside the text box itself (and hidden on mobile). Maybe the top right? Also, this method of copying won't work on all browsers so the button needs to be hidden on those that it doesn't.

Thanks for thinking of this!

sushain97:
@share-with-me, you're doing great work however remember that it's important that we wrap up work on one enhancement (i.e. merge the PR) before moving on to four others :)

share-with-me:
@sushain97 Yep, I realised that. I guess I am way too excited in trying to add features :P. I will first work wrap up the created PRs and then create new issues and PRs. :D

share-with-me:
Closing this PR and continuing it in a separate PR with the settings bar features.

Issue #122, opened 2017-03-22
Type Error occurs when nothing selected


share-with-me:
Currently, in Analyser or Generator mode, if we do not select anything from dropdown and click on `Analyse`, a type error of being `undefined` occurs on the console. Should this be eliminated by keeping a check on whether the dropdown item is selected and then letting the analyse function progress else put a console warn?  Or is this unnecessary from application point of view? 
I tried to implement this functionality and following are the screenshots:

![screen shot 2017-03-22 at 4 23 36 pm](https://cloud.githubusercontent.com/assets/8894636/24194766/563274ac-0f1d-11e7-87ad-acab4187161e.png)
![screen shot 2017-03-22 at 4 24 26 pm](https://cloud.githubusercontent.com/assets/8894636/24194775/5e5a3656-0f1d-11e7-81f0-8b4f3ad03c54.png)

sushain97:
Hm, I do want to prevent this error from occurring. However, there's no need for a console message. Send your PR and I'll take a look.

Thanks!

share-with-me:
@sushain97 I have created a PR for the same. Please have a look. #123

PR #123, opened 2017-03-22
TypeError fixed in analyse and generate modes

share-with-me:
This patch fixes the type error `undefined` that is displayed in the console when the user clicks on `Analyse` or `Generate` without selecting an item from the dropdown. This PR aims to solve issue #122 . 
Following are the screenshots:

**Before:** 
screen shot 2017-03-22 at 4 23 36 pm

**After:**
screen shot 2017-03-22 at 7 45 24 pm

The similar holds for Generate mode too.

### DIFF
--- a/assets/js/analyzer.js
+++ b/assets/js/analyzer.js
@@ -143,6 +143,10 @@ function populateSecondaryAnalyzerList() {
 }
 
 function analyze() {
+    if($('#primaryAnalyzerMode').val() == null) {
+        return;
+    }
+
     var analyzerMode = analyzers[$('#primaryAnalyzerMode').val()].length > 1
         ? $('#secondaryAnalyzerMode').val()
         : analyzers[$('#primaryAnalyzerMode').val()][0];
--- a/assets/js/analyzer.js
+++ b/assets/js/analyzer.js
@@ -143,7 +143,7 @@ function populateSecondaryAnalyzerList() {
 }
 
 function analyze() {
-    if($('#primaryAnalyzerMode').val() == null) {
+    if($('#primaryAnalyzerMode').val() === null) {
         return;
     }
 
--- a/assets/js/generator.js
+++ b/assets/js/generator.js
@@ -143,6 +143,10 @@ function populateSecondaryGeneratorList() {
 }
 
 function generate() {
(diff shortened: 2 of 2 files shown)


sushain97:
This problem would never occur if the APy instance had generators/analaysers since the list would be defaulted to the first one. If there are none, these modes shouldn't be activated. However, it is strange behavior this way so it's good to fix.

share-with-me:
> This problem would never occur if the APy instance had generators/analaysers since the list would be defaulted to the first one.
Umm, I did not completely understand what this means? I activated these modes through the `config` file.  Does it mean that the APy instance does not currently have generators/analysers? What do we mean by 'defaulting to first one'? 
Also, I have been trying to experiment with these modes but I am unable to fetch a language dropdown for the same. Do I have to separately mention the `langs` for these modes? Should I create a separate issue for this?
And and thanks for the merge :D . My first merge :P 
Also, I am still

sushain97:
> Does it mean that the APy instance does not currently have generators/analysers?

Yes. Those dropdowns should not be empty. Check out [turkic.apertium.org](turkic.apertium.org). You need to install generators or analysers on your machine for APy to get.

> Should I create a separate issue for this?

Nope.

> My first merge :P

Congrats! :) I would like one of your features to be merged in well before the GSoC proposal deadline.

share-with-me:
> Yes. Those dropdowns should not be empty. Check out turkic.apertium.org. You need to install generators or analysers on your machine for APy to get.

 I would install the generators and analysers from [turkic.apertium.org.](turkic.apertium.org). 

> Congrats! :) I would like one of your features to be merged in well before the GSoC proposal deadline.

Thank you so much! :D. Yes, I plan to wrap up 2 features, the button and the POST v/s GET request by today and then start with the proposal. Umm, I need to discuss the features that I have in mind and their doability, etc with you. Could have some time at around like 9 pm IST?

sushain97:
> Umm, I need to discuss the features that I have in mind and their doability, etc with you. Could have some time at around like 9 pm IST?

I will be in classes then. Shoot me an email ASAP (get the email from [skc.name](skc.name)) and I can respond to that sometime during the day.

share-with-me:
@sushain97 Thanks much! I would email you the ideas!

Issue #124, opened 2017-03-22
400-Bad Request Error on 'Detect Language'


share-with-me:
Currently, when the `#originalText` div is empty and we run the `Detect Language` feature, we get a `400-Bad Request Error`. It says the request is improperly configured and not conforming to the structure expected by API. Is it because the data is empty? I tried by keeping a check that the function works only when data is non-empty. And it gets resolved. Is my approach for the this issue correct?

**Before**
![screen shot 2017-03-22 at 11 25 57 pm](https://cloud.githubusercontent.com/assets/8894636/24213131/74c4fd7e-0f57-11e7-8a00-6878acea451e.png)

**After**
![screen shot 2017-03-22 at 11 41 20 pm](https://cloud.githubusercontent.com/assets/8894636/24213610/1ad8e562-0f59-11e7-999e-87d6fe63a418.png)

sushain97:
Preventing a request when the text is empty is probably wise. The console message is not necessary though.

Send a PR and I'll take a look. Thanks!

share-with-me:
@sushain97 I have created a PR for this #125 . Please have a look.
Thanks

share-with-me:
@sushain97  The PR has been merged. Should I close the issue?

PR #125, opened 2017-03-23
Removes the 400-Bad request error when detect language is clicked

share-with-me:
This patch first checks whether the data from `#originalText` is not empty. It lets the user make a `Detect Language` request only when the data is non-empty thereby, removing the `400-Bad Request Error.` This PR aims to solve the issue #124 . Following are the screenshots:

**Before**
screen shot 2017-03-23 at 9 05 56 am

**After**
screen shot 2017-03-23 at 9 06 27 am

### DIFF
--- a/assets/js/translator.js
+++ b/assets/js/translator.js
@@ -736,6 +736,10 @@ function downloadBrowserWarn() {
     }
 }
 function detectLanguage() {
+    if($('#originalText').val().length === 0) {
+        return;
+    }
+
     if(textTranslateRequest) {
         textTranslateRequest.abort();
     }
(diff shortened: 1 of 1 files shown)


assets/js/translator.js
```diff
@@ -736,6 +736,10 @@ function downloadBrowserWarn() {
     }
 }
 function detectLanguage() {
+    if($('#originalText').val().length === 0) {
```
sushain97: Extract `$('#originalText').val()` into a variable and replace other instances, perhaps `var originalText` since it's used later in this function as well.

share-with-me: @sushain97 Yepp, I would update the PR.

share-with-me:
@sushain97 I have updated the PR. Please have a look!
Thanks!

Issue #99, opened 2017-02-03
Website text right-to-left(-ish) if cookies are disabled


unhammer:
(Migrated from https://sourceforge.net/p/apertium/tickets/73/, originally reported by David Nemeskey)

I visited the Apertium website with my usual cookie blocker, and found the text is displayed incorrectly. For instance, on the About page, the `<li>` tags display the number on the right side, and are aligned right instead of left; the About button itself is also on the right side. The lower left label reads "!Spotted a mistake? Help us improve Apertium" (and links to the about page instead of e.g. the Sourceforge page, but that is beside the point).

If I enable cookies, the text is displayed correctly. While I expect most "normal" people to have cookies allowed by default, tech savvy people might not, and what they will see if they visit the site is not the best PR.

Firefox 38.0 on Linux Mint 17. I have NoScript installed, but I explicitly allowed scripts for the site.

_Kevin Brubeck Unhammer - 2015-06-21_

What cookie blocker? I tried various settings in Tor Browser, but can't reproduce it. Also, what exact URL was it? (The URL should change based on the language selected.)
     

_David Nemeskey - 2015-06-21_

URL: Just the default one: https://www.apertium.org/#translation. An extension to the bug report: I cannot even select a language without allowing cookies.

I am using Cookie Monster, version 1.2.5.1-signed.

I am attaching two screenshots that show how the site looks on my side.
     
![no_cookie_1](https://cloud.githubusercontent.com/assets/56868/22593249/f7347fb4-ea1d-11e6-8447-bbe910e676fe.png)
![no_cookie_2](https://cloud.githubusercontent.com/assets/56868/22593250/f75570c0-ea1d-11e6-9dab-4ee189a32688.png)

_Kevin Brubeck Unhammer - 2015-06-22_

I just tried adding Cookie Monster to Tor Browser, got the same version in about:addons.

I can't reproduce it with the default Cookie Monster settings, but if I change to "block all cookies" I get something looking like your screenshot. So the user-workaround is to temporarily allow cookies, but it'd be even better if we could fix it on the server :)

Also, my URL changes to https://www.apertium.org/index.eng.html?dir=oci-cat#translation when I load the page with cookies, but not when I block them.

share-with-me:
@unhammer @sushain97 I have attempted a patch to solve this in my [PR](https://github.com/goavki/apertium-html-tools/pull/129).

PR #129, opened 2017-04-02
Fixes the issue - Website appears right-to-left ish on disabling cookies

share-with-me:
This fix lets the JS script interact with the localStorage of the browser only when the cookies are enabled. Solves the issue #99 . A demo can be found [here](https://drive.google.com/open?id=0BwrSYlOsMCyWWUFzN2Nlci1QcXM)

### DIFF
--- a/assets/js/store.js
+++ b/assets/js/store.js
@@ -62,12 +62,26 @@ Store.prototype.has = function (key/*: string*/)/*: bool*/ {
 
 Store.prototype.able = function ()/*: bool*/ {
     try {
+        throwLocalStorageNotAccessibleError();
         return !!(window.localStorage);
     }
     catch(e) {
-        console.warn('Cookies are disabled. Thus, unable to access localStorage.');
-        return false;
+        if(e.name === 'localStorageNotAccessibleError') {
+            console.warn(e.message);
+            return false;
+        }
+        else {
+            console.warn('Some exception has occurred while loading the page.');
+            throw e;
+        }
     }
 };
+
+function throwLocalStorageNotAccessibleError() {
+    throw {
+        'name': 'localStorageNotAccessibleError',
+        'message': 'Cookies are disabled. Thus, unable to access localStorage.'
+    };
+}
 
 /*:: export {Store} */
--- a/assets/js/store.js
+++ b/assets/js/store.js
@@ -6,6 +6,8 @@
 }] */
 
 // eslint-disable-next-line func-style
+
+var isCookieEnabled = navigator.cookieEnabled;
 var Store = function (prefix/*: string*/)/*: void*/ {
     this.prefix = prefix;
 };
@@ -14,8 +16,10 @@ Store.prototype.get = function/*:: <T>*/ (key/*: string*/, fallback/*: T*/)/*: T
     if(fallback === undefined) {
         console.warn('Store.get with undefined fallback! Key:', key);
     }
-    if(!this.able()) {
-        return fallback;
+    if(isCookieEnabled === true) {
+        if(!this.able()) {
+            return fallback;
+        }
     }
     var fromStorage = window.localStorage[this.prefix + key];
     if(fromStorage === undefined) {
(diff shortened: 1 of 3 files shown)


sushain97:
What is the root cause? Why does having cookies disabled break the interface?

share-with-me:
@sushain97 Unable to find the root cause. My approach was, let the code interact with `store.js` and cache contents only when the cookies are enabled. I experimented, put a condition of checking if the cookies are enabled and it seemed to work! The site worked perfectly in the case when even if the user disables the cookies. It seems a solution is at hand but am still trying to get to the root cause.

sushain97:
> The site worked perfectly in the case when even if the user disables the cookies.

Unfortunately, that's impossible to determine since we don't have unit/integration tests :)

> It seems a solution is at hand but am still trying to get to the root cause.

Hm, okay. That's a good start but before merging any solution I would like to know why it fixes the problem. This way we can be certain that there are no inadvertent regressions.

share-with-me:
> Unfortunately, that's impossible to determine since we don't have unit/integration tests :)

Yes, indeed! My bad! I meant it did not break interface appearance wise! Yes, unit/integration tests are needed.

> Hm, okay. That's a good start but before merging any solution I would like to know why it fixes the problem. This way we can be certain that there are no inadvertent regressions.

Yes, I would keep trying to get to the root cause of it.
Thanks :D.

share-with-me:
After digging up a bit, this [stackoverflow link](https://stackoverflow.com/questions/26550770/html5-can-session-storage-local-storage-be-disabled-and-cookies-enabled), discussed about how localStorage works for different browsers in case the cookies are disabled. It mentions that for Chrome and FF, when we disable the cookies, the DOM Storage is automatically shut. However, for IE, it does remain open. 

The error mentioned for FF (in the stackoverflow link) is similar to the one I get in the console as in the following screenshot.
Does that mean that when we disable the cookies and try `Store.prototype.*` operations, the browsers raise an error and stop the page from loading further?!

I did not understand what `if(this.able()) { }` does in the `store.js` code. 

I believe the root cause is interacting with the `localStorage` of browsers in the event of cookies being disabled.

Please find the screenshot below.
screen shot 2017-06-06 at 9 53 16 am

sushain97:
> Does that mean that when we disable the cookies and try Store.prototype.* operations, the browsers raise an error and stop the page from loading further?!

Perhaps. What's on min.js L434?

> I did not understand what if(this.able()) { } does in the store.js code.

https://github.com/goavki/apertium-html-tools/blob/master/assets/js/store.js#L63-L65

checks whether LocalStorage is usable.

> I believe the root cause is interacting with the localStorage of browsers in the event of cookies being disabled.

If this is true, the proper solution is patching `Store.able` to me more accurate. It should be a < 5 line change.

unhammer:
> If this is true (which I agree with being the case on all but IE 11), the proper solution is patching Store.able to me more accurate. It should be a < 5 line change. However, since IE 11 cookies disabled != LocalStorage disabled, the ideal solution requires finding out how to figure out that LocalStorage is disabled reliably since the current Store.able implementation seems deficient.

If I understand correctly, I think we should avoid special-casing IE11 here. Just say that *Store requires both cookies and localstorage* to be `able`. (If Firefox and Chromium don't support localStorage when cookies are off, we shouldn't write special code to make IE do that).

unhammer:
> If this is true (which I agree with being the case on all but IE 11), the proper solution is patching Store.able to me more accurate. It should be a < 5 line change. However, since IE 11 cookies disabled != LocalStorage disabled, the ideal solution requires finding out how to figure out that LocalStorage is disabled reliably since the current Store.able implementation seems deficient.

If I understand correctly, I think we should avoid special-casing IE11 here. Just say that *Store requires both cookies and localstorage* to be `able`. This should require no changes to `persistence.js` or any other file than `store.js`. (If Firefox and Chromium don't support localStorage when cookies are off, we shouldn't write special code to make IE do that).

sushain97:
> If I understand correctly, I think we should avoid special-casing IE11 here. Just say that Store requires both cookies and localstorage to be able.

Hm, what if FF/Chrome changes the behavior of that checkbox to at some point not also disable LocalStorage? Would it not be more future-proof to reliably detect whether LocalStorage is available (which incidentally will make it work on IE11)?

share-with-me:
> If this is true (which I agree with being the case on all but IE 11), the proper solution is patching Store.able to me more accurate. It should be a < 5 line change. However, since IE 11 cookies disabled != LocalStorage disabled, the ideal solution requires finding out how to figure out that LocalStorage is disabled reliably since the current Store.able implementation seems deficient.

@sushain97  So if I am not wrong, we need to make `Store.able()` efficient enough to understand if the `localStorage` is accessible. Thus, if someone disabled the cookies, `localsStorage` automatically may shut off ( FF/Chrome ) and `Store.able()` will be able to find that out and prevent all the code inside `if( Store.able() ) { }`  from executing!

share-with-me:
@sushain97 , Please have a look! A try-catch logic to handle the same.

assets/js/store.js
```diff
@@ -61,7 +61,13 @@ Store.prototype.has = function (key/*: string*/)/*: bool*/ {
 };
 
 Store.prototype.able = function ()/*: bool*/ {
-    return !!(window.localStorage);
+    try {
+        return !!(window.localStorage);
+    }
+    catch(e) {
+        console.warn('Cookies are disabled. Thus, unable to access localStorage.');
```
sushain97: Is it possible to only handle the SecurityException? e.g. something like https://stackoverflow.com/questions/1433558/handling-specific-errors-in-javascript-think-exceptions

share-with-me: Yes, makes sense! On it.

share-with-me:
@sushain97 , Made the changes! Please review.

assets/js/store.js
```diff
@@ -62,12 +62,26 @@ Store.prototype.has = function (key/*: string*/)/*: bool*/ {
 
 Store.prototype.able = function ()/*: bool*/ {
     try {
+        throwLocalStorageNotAccessibleError();
```
sushain97: Doesn't this always throw the exception? I meant catching the `SecurityException` that you showed Firefox throws, not creating your own exception.

share-with-me: > Doesn't this always throw the exception? 

Yep. I tested it incorrectly. My bad!

>  I meant catching the SecurityException that you showed Firefox throws, not creating your own exception.

I will look for the error codes that exist for such exceptions! 
Thanks :D

share-with-me: We are looking at something like this right?!

```
    try {
        return !!(window.localStorage);
    }
    catch(e) {
        if(e.name === 'SecurityError') { // for FF, do similar for chrome too
            console.warn('Cookies disabled. Unable to access local browser storage.');
            return false;
        }
        else {
            throw e;
        }
    }

```

sushain97: Yes.

share-with-me: Perfect! It seems `SecurityError` is the name of the exception in FF. Looking for something similar in Chrome!

share-with-me:
@sushain97 Thank you so much :D. One more down :D

PR #134, opened 2017-05-31
Uses POST for larger contents and GET for smaller ones

share-with-me:
This PR aims to solve the issue #79. It uses **POST** method for larger texts and **GET** method for smaller texts. (The threshold length being 2048 chars). This functionality is currently for the **_translateText()_** function.


sushain97:
Don't forget this needs to be a comprehensive change that applies to all the modes (generator, analysis, etc.)

share-with-me:
@sushain97  Yes! I will do that!

assets/js/analyzer.js
```diff
@@ -1,5 +1,6 @@
 var analyzers = {}, analyzerData = {};
 var currentAnalyzerRequest;
+var THRESHOLD_REQUEST_LENGTH_FOR_ANALYZE = 2020;
```
sushain97: This variable should only be in one place. The current state of this code is a bit repetitive. You might want to move a method into `utils.js`

share-with-me: @sushain97 ,I initially planned to keep only a single `AJAX` function. Depending upon the endpoint I supply as the parameter, I would handle the `success` accordingly! And almost all of the rest of the parameters of the `AJAX` method would stay the same (which they are even now!). Do you think this is a good idea? A lot of code is repetitive in the current state.

sushain97: Sounds like a good approach overall. However, I would recommend against

> Depending upon the endpoint I supply as the parameter, I would handle the success accordingly!

Instead, pass in callbacks/extra options to the generalized function.

share-with-me: @sushain97 , Please have a look! I have modified it to a generalised function. I have added the functions to be used across scripts in the `/* global */` comment at the top. However, I believe I have not done that correctly yet!

assets/js/analyzer.js
```diff
@@ -157,81 +157,78 @@ function analyze() {
...
-                            else {
-                                joinedMorphemes[matches[k]] = [unit];
-                            }
-                        }
+    var request = {'lang': analyzerMode};
+    request.dataInput = $('#morphAnalyzerInput').val();
+    var methodType = (request.dataInput.length > THRESHOLD_REQUEST_LENGTH) ? 'POST' : 'GET';
+    var endpoint = '/analyze';
+    var callbackSuccess = handleAnalyzeSuccessResponse;
+    var callbackError = handleAnalyzeErrorResponse;
```
sushain97: There's no need to create these extra variables if you're going to only use them once.

assets/js/analyzer.js
```diff
@@ -157,81 +157,78 @@ function analyze() {
...
-                            if(joinedMorphemes[matches[k]]) {
-                                joinedMorphemes[matches[k]].push(unit);
-                            }
-                            else {
-                                joinedMorphemes[matches[k]] = [unit];
-                            }
-                        }
+    var request = {'lang': analyzerMode};
+    request.dataInput = $('#morphAnalyzerInput').val();
+    var methodType = (request.dataInput.length > THRESHOLD_REQUEST_LENGTH) ? 'POST' : 'GET';
```
sushain97: Is it possible to make this determination inside the generalized function?

assets/js/generator.js
```diff
@@ -158,31 +158,28 @@ function generate() {
...
+    var request = {'lang': generatorMode};
+    request.dataInput = $('#morphGeneratorInput').val();
+    var methodType = (request.dataInput.length > THRESHOLD_REQUEST_LENGTH) ? 'POST' : 'GET';
+    var endpoint = '/generate';
+    var callbackSuccess = handleGenerateSuccessResponse;
+    var callbackError = handleGenerateErrorResponse;
+    ajaxCallForAllModes(methodType, request, endpoint, callbackSuccess, callbackError, currentGeneratorRequest);
+}
+
+function handleGenerateSuccessResponse(data) {
```
sushain97: Nice work moving these into functions.

assets/js/util.js
```diff
@@ -242,8 +243,29 @@ function synchronizeTextareaHeights() {
...
 
-/*:: export {synchronizeTextareaHeights, modeEnabled, ajaxSend, ajaxComplete, filterLangList, onlyUnique,
-    SPACE_KEY_CODE, ENTER_KEY_CODE, HTTP_OK_CODE, HTTP_BAD_REQUEST_CODE, XHR_LOADING, XHR_DONE} */
+function ajaxCallForAllModes(methodType, request, endpoint, callbackSuccess, callbackError, requestName) {
+    requestName = $.ajax({
+        url: config.APY_URL + endpoint,
+        beforeSend: ajaxSend,
+        complete: function () {
+            ajaxComplete();
+            requestName = undefined;
```
sushain97: This doesn't do what you expect it to do since Javascript passes by value.

assets/js/util.js
```diff
@@ -242,8 +243,29 @@ function synchronizeTextareaHeights() {
     $('#translatedText').css('height', originalTextScrollHeight + 'px');
 }
 
-/*:: export {synchronizeTextareaHeights, modeEnabled, ajaxSend, ajaxComplete, filterLangList, onlyUnique,
-    SPACE_KEY_CODE, ENTER_KEY_CODE, HTTP_OK_CODE, HTTP_BAD_REQUEST_CODE, XHR_LOADING, XHR_DONE} */
+function ajaxCallForAllModes(methodType, request, endpoint, callbackSuccess, callbackError, requestName) {
```
sushain97: Name this function something like `apyCall`.

sushain97: Call it either `successCallback` or `success`, `callbackSuccess` is unconventional.

assets/js/translator.js
```diff
@@ -581,32 +581,31 @@ function translateText() {
...
+        $('#translatedText').val(data.responseData.translatedText);
+        $('#translatedText').removeClass('notAvailable text-danger');
+    }
+    else {
+        translationNotAvailable();
+    }
+}
+
+function handleTranslateErrorResponse(xOptions, error) {
+    translationNotAvailable();
```
sushain97: Can't you just pass in `translationNotAvailable`?

assets/js/util.js
```diff
@@ -242,8 +243,29 @@ function synchronizeTextareaHeights() {
...
+            ajaxComplete();
+            requestName = undefined;
+        },
+        data: request,
+        type: methodType,
+        contentType: 'application/x-www-form-urlencoded; charset=UTF-8',
+        dataType: 'json',
+        success: function (data) {
+            callbackSuccess(data);
+        },
```
sushain97: This can simply be `success: callbackSuccess` or even more elegantly: `success: success`. Instead of making each of these separate options passed through the arguments, you could make it accept an `options` Object and merge that Object with the one here. That would be the most extensible solution. Then the call sites for this function would provide their input as an object (`endpoint` can remain a separate argument).

assets/js/util.js
sushain97: This function probably needs to be added to `/* exported */`.

assets/js/util.js
share-with-me: Yep, I realised that now! Um, a workaround could be as mentioned in this [link](https://stackoverflow.com/questions/18178728/passing-a-global-variable-to-a-function). I would make use of `window[requestName]` to achieve the above functionality and pass the request names as a String argument. Inputs?

assets/js/generator.js
share-with-me: Thanks @sushain97  :D .

assets/js/util.js
sushain97: Yep. You'll have to use `window` :)

assets/js/analyzer.js
share-with-me: Um, when I create a request object for generator and analyzer modes, I make use of a JSON structure, and name my data variable as `q` ( similar to translator.js request object)! However, Hound throws an error of variable Identifier being too small ( atleast the length must be 4). Therefore, I renamed it to something like `dataInput`. I planned to make the check internal to the generalised function only ( as it is repetitive) but this is causing limitation. How do I resolve this?

assets/js/util.js
share-with-me: :D Thanks @sushain97 !

assets/js/translator.js
share-with-me: Yep! Makes sense! I wonder what I was thinking while writing this :P . Modified the code :D.

assets/js/util.js
share-with-me: > Instead of making each of these separate options passed through the arguments, you could make it accept an options Object and merge that Object with the one here. That would be the most extensible solution. Then the call sites for this function would provide their input as an object (endpoint can remain a separate argument).

@sushain97 , Um, do you mean something like this?

```
    var translateObject = {
        'methodType': methodType,
        'request': request,
        'success': success,
        'error': error,
        'requestName': 'textTranslateRequest'
    };
    apyCall(translateObject, endpoint);
```

And then the apyCall would look something like this?

```
function apyCall(requestObject, endpoint) {
    // code for checking the method type would be here instead of in the object
    textTranslateRequest = $.ajax({
        url: config.APY_URL + endpoint,
        beforeSend: ajaxSend,
        complete: function () {
            ajaxComplete();
            textTranslateRequest = undefined;
        },
        data: requestObject.request,
        type: requestObject.methodType,
        contentType: 'application/x-www-form-urlencoded; charset=UTF-8',
        dataType: 'json',
        success: requestObject.success,
        error: requestObject.error
    });
}

```

Here, each of the `success` and `error` callbacks have been defined before passing them into the request object. Finally, we pass this object as an argument to the `apyCall()`.

assets/js/generator.js
```diff
@@ -158,31 +158,30 @@ function generate() {
...
-            }
-            $('#morphGenOutput').removeClass('blurred');
-        },
-        error: function (xOptions, error) {
-            $('#morphGenOutput').text(error);
-            $('#morphGenOutput').removeClass('blurred');
-        }
-    });
+    var request = {'lang': generatorMode};
+    request.q = $('#morphGeneratorInput').val();
```
share-with-me: @sushain97 , This is the error I was talking about! :/

assets/js/util.js
```diff
@@ -242,7 +243,24 @@ function synchronizeTextareaHeights() {
...
+function apyCall(requestObject, endpoint) {
+    window[requestObject.requestName] = $.ajax({
+        url: config.APY_URL + endpoint,
+        beforeSend: ajaxSend,
+        complete: function () {
+            ajaxComplete();
+            window[requestObject.requestName] = undefined;
+        },
+        data: requestObject.request,
+        type: (requestObject.request.q.length > THRESHOLD_REQUEST_LENGTH) ? 'POST': 'GET',
```
share-with-me: @sushain97 , What does this error mean?

sushain97: http://eslint.org/docs/rules/space-infix-ops

assets/js/generator.js
sushain97: I'm pretty sure you can do `request['q'] = ...` to get around it.

share-with-me: Um, but why did a similar code with identifier as `q` worked well for translator.js?

assets/js/generator.js
```diff
@@ -158,31 +158,30 @@ function generate() {
...
-            }
-            $('#morphGenOutput').removeClass('blurred');
-        },
-        error: function (xOptions, error) {
-            $('#morphGenOutput').text(error);
-            $('#morphGenOutput').removeClass('blurred');
-        }
-    });
+    var request = {'lang': generatorMode};
+    request['q'] = $('#morphGeneratorInput').val();
```
share-with-me: @sushain97 , Um, `request['q']` throws this error?!

sushain97: Why not use an Object literal?

```
var request = {
    'lang': ...
    'q': ...
}
```

This is much better style anyway.

share-with-me: Umm yeah! And then accessing the variable inside the object through either `request['q']` or `request.q` should not give the error right?

sushain97: `request['q']` will probably still error in favor of `.q`. Don't worry too much about these kinds of errors if you can't fix them. We can always override the linter (you can Google for how).

share-with-me:
@sushain97 , Please review this PR so that I could wrap it up. Thanks!

assets/js/analyzer.js
```diff
@@ -3,7 +3,7 @@ var currentAnalyzerRequest;
 
 /* exported getAnalyzers */
 /* global config, modeEnabled, persistChoices, restoreChoices, localizeInterface, readCache, ajaxSend, ajaxComplete,
-    cache, getLangByCode, filterLangList, allowedLang, sendEvent */
+    cache, getLangByCode, filterLangList, allowedLang, sendEvent, apyCall */
```
sushain97: Could you call this `callApy`? The rest of the functions are named this way.

assets/js/analyzer.js
```diff
@@ -157,81 +157,82 @@ function analyze() {
...
+        'lang': analyzerMode,
+        'q': $('#morphAnalyzerInput').val()
+    };
+    var analyzeObject = {
+        'request': request,
+        'success': handleAnalyzeSuccessResponse,
+        'error': handleAnalyzeErrorResponse,
+        'requestName': 'currentAnalyzerRequest'
+    };
+    apyCall(analyzeObject, '/analyze');
```
sushain97: Why not inline everything? The variables are extraneous if you never use them again.

```
apyCall({
    'request': {
        'lang': analyzerMode,
        'q': $('#morphAnalyzerInput').val()
    },
    'success': handleAnalyzeSuccessResponse,
    'error': handleAnalyzeErrorResponse,
    'requestName': 'currentAnalyzerRequest'
}, '/analyze');
```

assets/js/generator.js
```diff
@@ -158,31 +158,32 @@ function generate() {
...
+        'lang': generatorMode,
+        'q': $('#morphGeneratorInput').val()
+    };
+    var generateObject = {
+        'request': request,
+        'success': handleGenerateSuccessResponse,
+        'error': handleGenerateErrorResponse,
+        'requestName': 'currentGeneratorRequest'
+    };
+    apyCall(generateObject, '/generate');
```
sushain97: Same inlining here.

assets/js/translator.js
```diff
@@ -846,7 +844,8 @@ function autoSelectDstLang() {
     }
 }
 
-/*:: import {synchronizeTextareaHeights, modeEnabled, ajaxSend, ajaxComplete, filterLangList, onlyUnique, getLangByCode} from "./util.js" */
+/*:: import {synchronizeTextareaHeights, modeEnabled, ajaxSend, ajaxComplete, filterLangList, onlyUnique, getLangByCode,
```
sushain97: Good catch modifying these as well!

assets/js/util.js
```diff
@@ -8,6 +8,7 @@ var SPACE_KEY_CODE = 32, ENTER_KEY_CODE = 13,
     XHR_LOADING = 3, XHR_DONE = 4;
 
 var TEXTAREA_AUTO_RESIZE_MINIMUM_WIDTH = 768;
+var THRESHOLD_REQUEST_LENGTH = 2000; // keep 48 charcters buffer for remaining params
```
sushain97:
```
var TEXTAREA_AUTO_RESIZE_MINIMUM_WIDTH = 768, 
    THRESHOLD_REQUEST_LENGTH = 2000; // ...
```

assets/js/util.js
```diff
@@ -242,8 +243,26 @@ function synchronizeTextareaHeights() {
...
-    SPACE_KEY_CODE, ENTER_KEY_CODE, HTTP_OK_CODE, HTTP_BAD_REQUEST_CODE, XHR_LOADING, XHR_DONE} */
+function apyCall(requestObject, endpoint) {
+    window[requestObject.requestName] = $.ajax({
+        url: config.APY_URL + endpoint,
+        beforeSend: ajaxSend,
+        complete: function () {
+            ajaxComplete();
+            window[requestObject.requestName] = undefined;
+        },
+        data: requestObject.request,
```
sushain97: Use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign or equivalent to make this extensible. You may have to polyfill this.

assets/js/util.js
```diff
@@ -242,8 +243,26 @@ function synchronizeTextareaHeights() {
     $('#translatedText').css('height', originalTextScrollHeight + 'px');
 }
 
-/*:: export {synchronizeTextareaHeights, modeEnabled, ajaxSend, ajaxComplete, filterLangList, onlyUnique,
-    SPACE_KEY_CODE, ENTER_KEY_CODE, HTTP_OK_CODE, HTTP_BAD_REQUEST_CODE, XHR_LOADING, XHR_DONE} */
+function apyCall(requestObject, endpoint) {
```
sushain97: `requestObject` -> `options`

assets/js/util.js
```diff
@@ -242,8 +243,26 @@ function synchronizeTextareaHeights() {
...
+    window[requestObject.requestName] = $.ajax({
```
sushain97: You may want to make `requestName` its own argument after using `Object.assign` as mentioned below since it isn't a parameter to `$.ajax`.

assets/js/analyzer.js
share-with-me: Yes, will make the change!

assets/js/analyzer.js
share-with-me: Perfect! Will modfy them :D

assets/js/translator.js
share-with-me: @sushain97  Thanks :D

assets/js/util.js
share-with-me: @sushain97 , I am not sure I understand what you mean by this point!  Make `requestName` as an argument? to what? Um, I believe I understand what the `Object.assign` does. But I am unable to comprehend what we are achieving here!

assets/js/util.js
share-with-me: `requestObject` -> `options`?  we are passing a JS object right!?

assets/js/util.js
sushain97: > Make requestName as an argument? to what?

The `apyCall` function.

> But I am unable to comprehend what we are achieving here!

With the `Object.assign` method, a call-site could include extra parameters to `$.ajax` without modifying the `apyCall` function itself.

assets/js/util.js
sushain97: I was suggesting an argument rename.

share-with-me: Hah! Will do that :D

assets/js/util.js
share-with-me: @sushain97 ! Um, 

> With the Object.assign method, a call-site could include extra parameters to $.ajax without modifying the apyCall function itself.

What would the `target` object hold in this case?  Also, why shouldn't I directly use `callApy(options, endpoint, requestName)` ? Should the `target` object be a JS object with `key`: `value` pairs corresponding to request names? 

( Not understanding how I have to use `Object.assign` function! ) A code snippet would be nice :D

sushain97: The `target` would be whichever arguments you're giving the `$.ajax` call that the `callApy` function creates on its own, e.g. `beforeSend`, `dataType` and others. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign#Merging_objects.

share-with-me: Um, okay. Apologies but I think I still don't have it clear. Could you explain your point with the code snippet below: ( or a code snippet to demonstrate your point would be perfect!)

```
var source =  {
        'data': request,
        'success': handleTranslateSuccessResponse,
        'error': translationNotAvailable,
        'requestName': 'textTranslateRequest'
}


var targetObj = {
	    url: config.APY_URL + endpoint,
        beforeSend: ajaxSend,
        requestName: 'requestName',
        complete: function () {
            ajaxComplete();
            window[options.requestName] = undefined;
        },
        data: options.request,
        type: (options.request.q.length > THRESHOLD_REQUEST_LENGTH) ? 'POST' : 'GET',
        contentType: 'application/x-www-form-urlencoded; charset=UTF-8',
        dataType: 'json',
        success: options.success,
        error: options.error
}

Object.assign(targetObj, source);

callApy(args?);


```

sushain97: The `Object.assign` call is within `callApy`. It should use the ajax arguments provided by the callee (`options` argument) which can be arbitrary (`source`) and combine those with its own default arguments (`target`).

share-with-me: Ohh! Um, something like the following:

```
function callApy(options, endpoint) {
    var targetObj = Object.assign({
        url: config.APY_URL + endpoint,
        beforeSend: ajaxSend,
        contentType: 'application/x-www-form-urlencoded; charset=UTF-8',
        dataType: 'json'
    },
    {
        data: options.request,
        type: (options.request.q.length > THRESHOLD_REQUEST_LENGTH) ? 'POST' : 'GET',
        complete: function () {
            ajaxComplete();
            window[options.requestName] = undefined;
        },
        success: options.success,
        error: options.error
    });

    window[options.requestName] = $.ajax(targetObj);
}
```

Here the parameters that are common across all the callees go into `target` by default. The remaining params go as a part of the `options` values in the `source`.

sushain97: Not exactly. The whole point of having an arbitrary `options` Object and using `Object.assign` is that the callee can pass in arguments `callApy` doesn't know. The `source` to `Object.assign` should simply be `options`.

share-with-me: > The whole point of having an arbitrary options Object and using Object.assign is that the callee can pass in arguments callApy doesn't know.

What does arbitrary mean? It can have any random `key: value` pairs.
 So, it may happen that, my `options` object is simply a `{a: 1, b: 2}` and I may pass this to `callApy` as `options` argument? In this case, how do I set the `data`, `success` etc values in the `ajax` call?

One solution I thought of is to have my `options` object modified as the one in below snippet. I make use of the `keys` that are valid in an ajax call. Now, when `Object.assign` runs, it copies the contents of this object into the target and executes the function! 

```
callApy({
       'data': request,
       'success': handleTranslateSuccessResponse,
       'error': translationNotAvailable,
       'requestName': 'textTranslateRequest'
}, endpoint);

function callApy(options, endpoint) {
    var targetObj = Object.assign({
        url: config.APY_URL + endpoint,
        beforeSend: ajaxSend,
        contentType: 'application/x-www-form-urlencoded; charset=UTF-8',
        dataType: 'json'
    }, options);

    window[options.requestName] = $.ajax(targetObj);
}

```

sushain97: Yep, this is exactly what I've been trying to lead you towards :)

sushain97: No need to define an extra variable, just pass it directly to `$.ajax` but this looks good otherwise.

share-with-me: woah finally :D . Um, for the method `type`, in this case, should I check it in my `options` object itself before passing it into the `callApy()` ? Otherwise, it would require modifying the `targetObj`!

sushain97: > Otherwise, it would require modifying the targetObj!

Do it this way. It makes sense to have this logic inside `callApy`.

share-with-me: Ohh cool! On it :D. Yayy!!

share-with-me: Um, just one last issue! Would it make sense to pass `requestName` as an argument to `callApy` itself? I could write the `complete: function() { }` inside the target object itself directly then!

sushain97: > Would it make sense to pass requestName as an argument to callApy itself?

Yes; I think I mentioned this earlier.

> I could write the complete: function() { } inside the target object itself directly then!

I think we should leave that to the caller.

share-with-me: > Yes; I think I mentioned this earlier.

I now get what you were saying since then :D.

> I think we should leave that to the caller.

Meaning it should be a part of `options`?

sushain97: > Meaning it should be a part of options?

Yep.

share-with-me: :D On it

assets/js/util.js
```diff
@@ -242,8 +243,21 @@ function synchronizeTextareaHeights() {
...
+function callApy(options, endpoint, requestName) {
+    var callApyObject = Object.assign({
+        url: config.APY_URL + endpoint,
+        beforeSend: ajaxSend,
+        contentType: 'application/x-www-form-urlencoded; charset=UTF-8',
+        dataType: 'json'
+    }, options);
+
+    callApyObject.type = (callApyObject.data.q.length > THRESHOLD_REQUEST_LENGTH) ? 'POST' : 'GET';
+    window[requestName] = $.ajax(callApyObject);
```
sushain97: Actually, can we do something like this:

```
 +  return $.ajax(Object.assign({
 +        url: config.APY_URL + endpoint,
 +        beforeSend: ajaxSend,
 +        contentType: 'application/x-www-form-urlencoded; charset=UTF-8',
 +        dataType: 'json',
 +        type: (callApyObject.data.q.length > THRESHOLD_REQUEST_LENGTH) ? 'POST' : 'GET',
 +    }, options));
```
and then handle the `requestName = ...` at the callsite?

share-with-me: Actually, yeah! It makes sense :D. Um, how will `type` field get populated this way? If we are eliminating `callApyObject`, then we should use:

`type: (options.data.q..........);` right?

assets/js/util.js
```diff
@@ -242,8 +243,19 @@ function synchronizeTextareaHeights() {
...
 
-/*:: export {synchronizeTextareaHeights, modeEnabled, ajaxSend, ajaxComplete, filterLangList, onlyUnique,
-    SPACE_KEY_CODE, ENTER_KEY_CODE, HTTP_OK_CODE, HTTP_BAD_REQUEST_CODE, XHR_LOADING, XHR_DONE} */
+function callApy(options, endpoint) {
+    return $.ajax(Object.assign({
+        url: config.APY_URL + endpoint,
+        beforeSend: ajaxSend,
+        contentType: 'application/x-www-form-urlencoded; charset=UTF-8',
+        dataType: 'json',
+        type: (options.data.q.length > THRESHOLD_REQUEST_LENGTH) ? 'POST' : 'GET'
```
sushain97: This should use `$.jsonp` instead of a GET request since CORS support is not uniform.

share-with-me: Um, `$.jsonp` instead of `$.ajax` you mean? `$.jsonp` wraps around `GET` request and not the `POST`.

sushain97: Instead of `GET`, yes.

share-with-me: You are saying something like this?:

```
type: (options.data.q.length > THRESHOLD_REQUEST_LENGTH) ? 'POST' : 'JSONP'
```

sushain97: No, that does not work. You need to switch between JSONP and POST requests.

share-with-me: Um , something like below would fail too? Under CORS?

```
dataType: (options.data.q.length > THRESHOLD_REQUEST_LENGTH) ? 'JSON' : 'JSONP',
type: (options.data.q.length > THRESHOLD_REQUEST_LENGTH) ? 'POST' : 'GET'
```

sushain97: Look into JSONP and CORS. You need both `$.get`/`$.ajax` and `$.jsonp`.

share-with-me: Yep, on it!

share-with-me: @sushain97 , I read through what `jsonp` means and why it works only `GET`. Are you suggesting something like the following?

```
function callApy(options, endpoint) {
    var callApyObject = Object.assign({
        url: config.APY_URL + endpoint,
        beforeSend: ajaxSend,
        contentType: 'application/x-www-form-urlencoded; charset=UTF-8'
    }, options);

    if(options.data.q.length > THRESHOLD_REQUEST_LENGTH) {
        callApyObject.type = 'POST';
        callApyObject.dataType = 'json';
        return $.ajax(callApyObject);
    }
    else {
        callApyObject.type = 'GET';
        callApyObject.dataType = 'jsonp';
        return $.jsonp(callApyObject);
    }
}
```

sushain97: Yep. Rename `callApyObject` to `requestOptions` or something. Be sure to ensure it works by testing!

sushain97: Instead of using `options.data.q.length`, could you use the sum of the lengths of all keys and values + the 1 * number of those (for the "+" symbol) + length of full url? This way could possibly break with more complex endpoints or those that don't use `q`.

share-with-me: Yep, makes sense! I would modify the query length!

share-with-me: > 1 * number of those (for the "+" symbol) 

@sushain97 , this part, umm, the `encodeURIComponent` function does produce kind of `= & ?` symbols too right? When we say the limitation on the `GET` request length, it is actually in terms of the URI length or the elements in `q` data packet?

share-with-me: > 1 * number of those (for the "+" symbol)

This is for the value in the `originalText` right? As in I find out `encodeURIComponent($('#originalText').val()).length`? This will have the `+` in between strings of the content?

sushain97: Sorry, I meant for `&`. However, you are correct `=` needs to be included as well. The most ideal solution is find a (simple) way to actually construct the URL and check.

share-with-me: If by some means, I am able to check the length of the URL with `&q= 'encodeURIdata'` and its length, we are done? 

Reading up on the web states that it is the `URL` length which imposes limitations. If that is the case, traversing the keys and values and summing up their lengths to obtain the net length would not be what we want right?

sushain97: > Reading up on the web states that it is the URL length which imposes limitations. If that is the case, traversing the keys and values and summing up their lengths to obtain the net length would not be what we want right?

That's not correct. The URL length includes those data parameter key value pairs. Take a look at how an APY request looks in the Network panel of developer tools.

share-with-me: >  The most ideal solution is find a (simple) way to actually construct the URL and check.

&

> The URL length includes those data parameter key value pairs.

Okay, so the most ideal way would be:
1. Sum up the lengths of keys and values of all the params in `requestOptions` object.
2. Construct the URL similar to something that this snippet from `persistChoices` function generates?

```
        $.each(urlParamNames, function () {
            var urlParam = getURLParam(this);
            if(urlParam) {
                urlParams.push(this + '=' + encodeURIComponent(urlParam));
            }
        });

//

        else if(getURLParam('q' + qName).length > 0) {
            urlParams.push('q' + qName + '=' + getURLParam('q' + qName));
        }

        var newURL = window.location.pathname + (urlParams.length > 0 ? '?' + urlParams.join('&') : '') + hash;


```

sushain97: > Sum up the lengths of keys and values of all the params in requestOptions object.

Not exactly. Only `options.data` gets sent in the URL.

> I am unable to determine ultimately what all parameters will contribute to the GET request packet!

Read into `$.get` and the GET HTTP request.

> Construct the URL similar to something that this snippet from persistChoices function generates? and include what parts of this URL as a part of my request data length? (Eg: only & and =, + kind of stuff? Since the other values have already been considered while traversing the object

I would suspect that there's an easier method with jQuery. If not, then yes, a similar fashion would suffice.

share-with-me: > Read into $.get and the GET HTTP request.

Yep, read into :)

> I would suspect that there's an easier method with jQuery. If not, then yes, a similar fashion would suffice.

`$.param()` might come to rescue! But my `URL` will look like the `window.location.href` appended by `window.location.pathname` followed by the `q= 'encodeddata'` which is nothing but the contents of `originalText`? And it the length of this URL that I want?

sushain97: > $.param() might come to rescue!

Looks promising.

> followed by the q= 'encodeddata' which is nothing but the contents of originalText?

There are other parameters. This should be built in a way that is extensible for other modes as well. Don't just think about the translation mode! We want to build things that generalize well.

> And it the length of this URL that I want?

Yes. The limiting factor is the size of the URL.

share-with-me: For instance, I ran this example with `$.param()` and `originalText` content as `hello world`. The URL that gets generated is as follows: (I modified `window.location.href` as it gets modified according to the contents of `persistChoices` function for implementation purpose)

This is the output I get: `http://localhost:8080/index.eng.html?langpair=eng%7Cspa&q=hello+world&markUnknown=yes`. I could check the length of this URL and make the decision right?

sushain97: > I could check the length of this URL and make the decision right?

Yep.

assets/js/util.js
```diff
@@ -1,13 +1,14 @@
...
 /* exported SPACE_KEY_CODE, ENTER_KEY_CODE, HTTP_OK_CODE, HTTP_BAD_REQUEST_CODE, XHR_LOADING, XHR_DONE */
 /* global config, persistChoices, iso639Codes, iso639CodesInverse */
 
 var SPACE_KEY_CODE = 32, ENTER_KEY_CODE = 13,
     HTTP_OK_CODE = 200, HTTP_BAD_REQUEST_CODE = 400,
     XHR_LOADING = 3, XHR_DONE = 4;
 
-var TEXTAREA_AUTO_RESIZE_MINIMUM_WIDTH = 768;
+var TEXTAREA_AUTO_RESIZE_MINIMUM_WIDTH = 768,
+    THRESHOLD_REQUEST_LENGTH = 2000; // keep 48 charcters buffer for remaining params
```
sushain97: You might be able to modify this back to 2048 since we've taken into account other parameters.

assets/js/util.js
share-with-me: On it!

assets/js/util.js
share-with-me: Yes

Issue #65, opened 2016-06-08
dropdown button sometimes disappears


jonorthwash:
The dropdown button for the list of languages on translation input side is sometimes covered by the reverse button:
![screenshot_2016-06-08_183645](https://cloud.githubusercontent.com/assets/963849/15913464/2406133c-2da8-11e6-9109-4509946f9ff5.png)

sushain97:
Having trouble duplicating this one as well. Browser?

jonorthwash:
Chromium.  In order to reproduce, you need to select languages that take up the right amount of space.  Here's another instance of it.
![screenshot_2016-06-25_133939](https://cloud.githubusercontent.com/assets/963849/16358179/4f233aa6-3ada-11e6-9391-f470262d89d9.png)

SilentFlame:
unable to reproduce.

I guess because of small size language names present for English language based browsers.
How can I reproduce the issue?

jonorthwash:
@SilentFlame, you can set the language to Kyrgyz (кыргызча) or Russian (русский) using the localisation dropdown in the top right, as in the examples.  I think you just have to cycle through the various available pairs until it lines up like this.

jonorthwash:
I was just able to reproduce it in firefox by getting long language names all to line up on the right side.
![selection_004](https://cloud.githubusercontent.com/assets/963849/20770337/5bc10e22-b713-11e6-93b0-b2564505b0a2.png)

jonorthwash:
@SilentFlame, it also seems to help to make the browser window narrower, so that it switches to one of the smaller display modes.  Then you don't need such long language names to trigger the issue.

share-with-me:
Yep! I could reproduce this problem on chrome. Narrowing the browser either lets a `language` button overlap the `reverse` button or vice-versa. Following are the screenshots. I plan to solve this issue.

![screen shot 2017-03-26 at 10 10 35 pm](https://cloud.githubusercontent.com/assets/8894636/24333257/61192688-1272-11e7-8f21-8e78e7073b05.png)

![screen shot 2017-03-26 at 10 13 51 pm](https://cloud.githubusercontent.com/assets/8894636/24333259/697b143a-1272-11e7-9c47-c1a6be4ee31b.png)

PR #141, opened 2017-06-17
Fixes the issue of dropdown button being overlapped

share-with-me:
Fixes issue #65 .
The fix ensures that there is no overlap between `swapButton` and the `srcLangSelectors` divs in any case.

### DIFF
--- a/assets/css/style.css
+++ b/assets/css/style.css
@@ -249,6 +249,7 @@ html[dir=rtl] #srcLanguages {
 
 #srcLangSelectors {
     z-index: 100;
+    width: 90%;
 }
 
 @media(min-width: 768px) {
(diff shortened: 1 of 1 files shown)


assets/css/style.css
```diff
@@ -249,6 +249,7 @@ html[dir=rtl] #srcLanguages {
 
 #srcLangSelectors {
     z-index: 100;
+    width: 90%;
```
sushain97: Can you use `calc` with the size of the swap button subtracted instead?

share-with-me: Yes, here the width of swap button will be constant right? Like I can directly hardcode the value of `swapLangBtn` width in `calc` function? Something like below: ( 35px is the width I am subtracting )

`width: calc(100% - 35px);`

sushain97:
Yep, it will be constant. Use whatever the dev tools say it is.

sushain97:
Nice work!

share-with-me:
Thanks @sushain97  :D

PR #153, opened 2017-06-20
Unobtrusive feature

share-with-me:


share-with-me:
@sushain97 , `newline-per-chained-call` means something like the following?? What about indentation? Can I put a new line at each function call?

```
$('#installationNotice').removeClass('hide').fadeIn('slow').
                                                     delay(messageDuration).
                                                     fadeOut('slow', hideUnobtrusiveWarning);
```

share-with-me:
Um, @sushain97 , Could you review my above comment on this? So i could get Travis pass the code!

sushain97:
> @sushain97 , newline-per-chained-call means something like the following?? What about indentation? Can I put a new line at each function call?

Each chained call goes on a separate line with the dots at the start of the line.

All the rules are documented on ESLint, e.g. http://eslint.org/docs/rules/newline-per-chained-call.

share-with-me:
@sushain97 Thanks for the link! Um, I need to fix the warnings too? :P . Errors should do right? :P

sushain97:
> @sushain97 Thanks for the link! Um, I need to fix the warnings too? :P . Errors should do right? :P

Nope. I would make all the warnings errors as well but that is tricky to do with `sass-lint` as far as I can tell. No warnings or errors ==> mergeable.

assets/strings/eng.json
```diff
@@ -61,5 +61,7 @@
...
+    "Contact_Para": "<h4>IRC channel</h4> <p>The quickest way to contact us is by joining our <a href='http://wiki.apertium.org/wiki/IRC' target='_blank'><abbr title='Internet Relay Chat'>IRC</abbr> channel, #apertium</a> at irc.freenode.net, where users and developers of Apertium meet. You don't need an IRC client; you can use <a href='http://webchat.freenode.net/?channels=apertium'>freenode webchat</a>. </p> <h4>Mailing list</h4> <p>Also, subscribe to the <a href='https://lists.sourceforge.net/lists/listinfo/apertium-stuff' target='_blank'> apertium-stuff mailing list</a>, where you can post longer proposals or issues, as well as follow general Apertium discussions. </p> <h4>Contact</h4> <p>Feel free to contact us via the <a href='mailto:apertium-contact@lists.sourceforge.net' target='_blank'>apertium-contact mailing list</a> if you find a mistake, there's a project you would like to see us work on, or you would like to help out.</p>",
+    "Installation_Notice_Title": "Inconvenience",
```
sushain97: `Inconvenience` => `Install Apertium`

assets/strings/eng.json
```diff
@@ -61,5 +61,7 @@
...
+    "Installation_Notice": "We are highly sorry for the inconvenience caused but our server appears loaded now. In order to set apertium locally, please refer to this <a href='http://wiki.apertium.org/wiki/Apertium-apy' target='_blank'> link</a>."
```
sushain97: `"Are you experiencing slow responses? Our servers might be overloaded. Learn how to <a href="http://wiki.apertium.org/wiki/Installation">install Apertium locally</a>."`

assets/js/util.js
```diff
@@ -293,13 +300,82 @@ function callApy(options, endpoint) {
...
     }
     return $.jsonp(requestOptions);
 }
 
+function checkServiceLoadTimes(requestDuration) {
+    var lastNRequestsDuration = [];
+    var storedLastNRequestsDuration = [];
+    var demoThreshold = 10000; // for demo purposes only
+    try {
+        window.sessionStorage;
```
sushain97: I thought about this a bit more and I don't think these need to be stored in `sessionStorage`. 

From our conversation on the PR:

> My current understanding is, for the current session ie until user closes the tab/window, I store all the page load times in a session variable and then calculate the average?

This is fine. Simple and easy.

From this I assumed that session storage persisted only through the life of the tab but it seems to be the life of the browser instead. I think it makes more sense (and drastically simplifies the code) for this to just be per page. We should also make sure to only display the message to the user once. We don't want them to keep seeing it.

assets/js/util.js
```diff
@@ -11,6 +11,11 @@ var TEXTAREA_AUTO_RESIZE_MINIMUM_WIDTH = 768,
     BACK_TO_TOP_BUTTON_ACTIVATION_HEIGHT = 300,
     THRESHOLD_REQUEST_URL_LENGTH = 2000; // maintain 48 characters buffer for generated parameters
 
+var apyRequestStartTime,
+    INSTALLATION_NOTIFICATION_REQUESTS_BUFFER_LENGTH = 10,
+    INDIVIDUAL_DURATION_THRESHOLD = 2000,
```
sushain97: Prefix these two variables with `INSTALLATION_NOTIFICATION_`

assets/js/util.js
share-with-me: > From this I assumed that session storage persisted only through the life of the tab but it seems to be the life of the browser instead.

Yes, a bit more reading makes me realize this!

>  I think it makes more sense (and drastically simplifies the code) for this to just be per page.

Oh! Um, that means I could just store these values in an `array`. And everytime the user refreshes the page, all of this starts afresh? 

> We should also make sure to only display the message to the user once. We don't want them to keep seeing it.

Yes, we do not want to annoy them!!

share-with-me:
> (travis and conflict)

On it

share-with-me:
@sushain97 , Um, `fa-times` is a `font-awesome` class. Travis is giving an error stating it should be made camelcase. Should I also add a custom class there and use that to style the element?

sushain97:
> And then style using faTimes in CSS?

That's a good approach. An easier solution is to simply use the linter ignore syntax (look for it in other places in the file).

share-with-me:
>  An easier solution is to simply use the linter ignore syntax (look for it in other places in the file)

Thanks @sushain97 :D. Done!!!

assets/js/util.js
```diff
@@ -11,6 +11,12 @@ var TEXTAREA_AUTO_RESIZE_MINIMUM_WIDTH = 768,
     BACK_TO_TOP_BUTTON_ACTIVATION_HEIGHT = 300,
     APY_REQUEST_URL_THRESHOLD_LENGTH = 2000; // maintain 48 characters buffer for generated parameters
 
+var apyRequestStartTime, installationNoticeFlag = true, lastNRequestsDuration = [],
+    requestsMade = 0, cumulativeRequestsTime = 0,
+    INSTALLATION_NOTIFICATION_REQUESTS_BUFFER_LENGTH = 10,
```
sushain97: Move the constants to the place where the rest of the constants are.

assets/js/util.js
```diff
@@ -11,6 +11,12 @@ var TEXTAREA_AUTO_RESIZE_MINIMUM_WIDTH = 768,
     BACK_TO_TOP_BUTTON_ACTIVATION_HEIGHT = 300,
     APY_REQUEST_URL_THRESHOLD_LENGTH = 2000; // maintain 48 characters buffer for generated parameters
 
+var apyRequestStartTime, installationNoticeFlag = true, lastNRequestsDuration = [],
```
sushain97: Use more descriptive names such as `installationNoticeShown`.

assets/js/util.js
```diff
@@ -162,6 +171,7 @@ $(document).ready(function () {
         return false;
     });
 
+    $('#installationNotice').addClass('hide');
```
sushain97: Why not just `.hide()`?

sushain97:
Trying out this feature, I'd also like the notification to appear if a running request takes longer than the `INSTALLATION_NOTIFICATION_INDIVIDUAL_DURATION_THRESHOLD`. Also, we're switching around with our terminology. Should we use notice or notification...

sushain97:
For the notification styling, you're on a good track. Try something a little closer to vanilla Bootstrap: http://getbootstrap.com/components/#alerts. Also, the link needs to be `target="_blank"`.

share-with-me:
> Also, we're switching around with our terminology. Should we use notice or notification...

According to me, `notification` makes more sense!

assets/js/util.js
share-with-me: Yes!

assets/js/util.js
share-with-me: Done

share-with-me:
> Trying out this feature, I'd also like the notification to appear if a running request takes longer than the INSTALLATION_NOTIFICATION_INDIVIDUAL_DURATION_THRESHOLD.

Makes sense! Modified the code :)

> Why not just .hide()?

I have added classes `hidden-xs hidden-sm` to HTML. This adds a property of `display: block !important` (for larger screen devices). Thus, simply using `hide()` or `show()` does not override these bootstrap inline styles. Adding and removing `hide` class overrides these limitations and displays and shows the `div`. Thus, I make use of `addClass('hide')` and `removeClass('hide')`. This [stackoverflow link](https://stackoverflow.com/questions/18568736/how-to-hide-element-using-twitter-bootstrap-3-and-show-it-using-jquery) discusses about how `hide, addClass and removeClass` can be used to achieve the desired functionality.

share-with-me:
>  Try something a little closer to vanilla Bootstrap: http://getbootstrap.com/components/#alerts.

Um, I have used `alert-warning`. Should I make use of `alert-info/success`?

sushain97:
> Um, I have used alert-warning. Should I make use of alert-info/success?

Hm, either `warning` or `info`... Show me how both look?

share-with-me:
> Hm, either warning or info... Show me how both look?

Yes :D.

1. Warning:

screenshot from 2017-06-25 00-26-21

2. Info:

screenshot from 2017-06-25 00-27-51

3. Success:

screenshot from 2017-06-25 00-28-31

sushain97:
I'm honestly fine with either of the first two. Is there a reason you're using a large white border? Its seems a bit unsightly to me.

share-with-me:
> Is there a reason you're using a large white border?

Um, no! I just did not change the notification CSS :D. Have removed it now.

assets/js/util.js
```diff
@@ -9,7 +9,14 @@ var SPACE_KEY_CODE = 32, ENTER_KEY_CODE = 13,
 
 var TEXTAREA_AUTO_RESIZE_MINIMUM_WIDTH = 768,
     BACK_TO_TOP_BUTTON_ACTIVATION_HEIGHT = 300,
-    APY_REQUEST_URL_THRESHOLD_LENGTH = 2000; // maintain 48 characters buffer for generated parameters
+    APY_REQUEST_URL_THRESHOLD_LENGTH = 2000, // maintain 48 characters buffer for generated parameters
+    INSTALLATION_NOTIFICATION_REQUESTS_BUFFER_LENGTH = 10,
+    INSTALLATION_NOTIFICATION_INDIVIDUAL_DURATION_THRESHOLD = 2000,
```
sushain97: double these values

assets/js/util.js
```diff
@@ -9,7 +9,14 @@ var SPACE_KEY_CODE = 32, ENTER_KEY_CODE = 13,
...
+    INSTALLATION_NOTIFICATION_CUMULATIVE_DURATION_THRESHOLD = 1500;
+
+var apyRequestStartTime, installationNotificationShown = false, isCallApyCompleted,
```
sushain97: `isCallApyCompleted` -> `apyRequestTimeout`

assets/js/util.js
```diff
@@ -45,6 +52,10 @@ function ajaxSend() {
 
 function ajaxComplete() {
     $('#loadingIndicator').hide();
+    if(installationNotificationShown === false) {
```
sushain97: `!installationNotificationShown`

assets/js/util.js
```diff
@@ -293,13 +305,63 @@ function callApy(options, endpoint) {
...
+    var averageRequestsDuration = cumulativeRequestsTime / requestsMade;
+
+    if(requestDuration > INSTALLATION_NOTIFICATION_INDIVIDUAL_DURATION_THRESHOLD ||
+        averageRequestsDuration > INSTALLATION_NOTIFICATION_CUMULATIVE_DURATION_THRESHOLD) {
+        displayUnobtrusiveWarning();
+    }
+}
+
+function displayUnobtrusiveWarning() {
+    var messageDuration = 10000;
```
sushain97: Move this to a constant.

assets/js/util.js
```diff
@@ -9,7 +9,14 @@ var SPACE_KEY_CODE = 32, ENTER_KEY_CODE = 13,
...
 var TEXTAREA_AUTO_RESIZE_MINIMUM_WIDTH = 768,
     BACK_TO_TOP_BUTTON_ACTIVATION_HEIGHT = 300,
-    APY_REQUEST_URL_THRESHOLD_LENGTH = 2000; // maintain 48 characters buffer for generated parameters
+    APY_REQUEST_URL_THRESHOLD_LENGTH = 2000, // maintain 48 characters buffer for generated parameters
+    INSTALLATION_NOTIFICATION_REQUESTS_BUFFER_LENGTH = 10,
+    INSTALLATION_NOTIFICATION_INDIVIDUAL_DURATION_THRESHOLD = 4000,
+    INSTALLATION_NOTIFICATION_CUMULATIVE_DURATION_THRESHOLD = 3000;
+
+var apyRequestStartTime, installationNotificationShown = false, apyRequestTimeout,
+    lastNRequestsDuration = [], requestsMade = 0, cumulativeRequestsTime = 0, installationNotificationDuration = 10000;
```
sushain97: This constant needs to be moved up as well.

assets/js/util.js
```diff
@@ -293,13 +305,62 @@ function callApy(options, endpoint) {
     var requestUrl = window.location.protocol + window.location.hostname +
         window.location.pathname + '?' + $.param(requestOptions.data);
 
+    if(!installationNotificationShown) {
```
sushain97: This is always going to be `true` when the page loads.

assets/js/util.js
```diff
@@ -293,13 +305,62 @@ function callApy(options, endpoint) {
...
+        apyRequestStartTime = Date.now();
+        apyRequestTimeout = setTimeout(function () {
+            displayUnobtrusiveWarning();
```
sushain97: Isn't this going to run regardless of how long the request takes?

assets/js/util.js
share-with-me: Yes, because the notification is not yet shown! Are you suggesting that the value holds even when the user refreshes?

assets/js/util.js
share-with-me: Nope, I am clearing the `apyRequestTimeout` variable on `ajaxComplete`. Thus, if my AJAX request is completed before the INDIVIDUAL_THRESHOLD, the variable will be cleared and it will not execute.

assets/js/util.js
sushain97: Whoops, I mistook the context in which this gets run. Can we move this check into `displayUnobtrusiveWarning` (which we should rename to `displayInstallationNotification`) - you can make it an early return. Then it only needs to happen in one place.

assets/js/util.js
sushain97: Wow, I completely missed that. I think it's time to go to bed!

assets/js/util.js
```diff
@@ -293,13 +306,62 @@ function callApy(options, endpoint) {
...
+    $('#installationNotice').mouseover(function () {
+        $(this).stop(true);
+    }).mouseout(function () {
+        $(this).animate()
+            .delay(INSTALLATION_NOTIFICATION_DURATION)
+            .fadeOut('slow', hideUnobtrusiveWarning);
+    });
+}
+
+function hideUnobtrusiveWarning() {
```
sushain97: (also change the name of this one)

assets/js/util.js
share-with-me: Hahahahahahhaha! yes, please sleep well :P

assets/js/util.js
share-with-me: On it :D

share-with-me:
Done @sushain97  :D

PR #156, opened 2017-06-22
Fix container animation issue - issue 149

### LINKED ISSUE #149: Containers can all disappear
if you click on one mode in the navbar and then on the original while it's transitioning:

image

share-with-me:
Attempt to fix #149 . I add a `stop()` method to the `hide()` and `show()` methods on clicking a container nav option. This ensures that when a nav `mode` is clicked, all animations applied on the navbar modes stop and continue afresh. According to me, currently, as the animations are not queued up( as `queue: false` ), clicking on another `mode` whilst a transitional animation is going on kind of messes up with the animations. All the records of animations get vanished thereby resulting in a blank screen.

### DIFF
--- a/assets/js/util.js
+++ b/assets/js/util.js
@@ -108,10 +108,10 @@ $(document).ready(function () {
         var mode = $(this).data('mode');
         $('.nav li').removeClass('active');
         $(this).parent('li').addClass('active');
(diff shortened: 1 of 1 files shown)


share-with-me:
Thanks @sushain97  :D

Issue #161, opened 2017-07-01
TranslateDoc() called on page refresh in analyze/generate modes


share-with-me:
Currently, if we switch to `Morphological Analysis/Generation` tab interface and then refresh, `translateDoc()` gets called which shouldn't.

![screenshot from 2017-07-02 00-09-48](https://user-images.githubusercontent.com/8894636/27764602-cb504b14-5eba-11e7-8efe-15960f0681f9.png)

PR #162, opened 2017-07-03
Call translateDoc() only when the div is visible - Fix issue 161

share-with-me:
#161 fixed.

### DIFF
--- a/assets/js/translator.js
+++ b/assets/js/translator.js
@@ -562,7 +562,7 @@ function translate() {
     if($('div#translateText').is(':visible')) {
         translateText();
     }
-    else {
+    else if($('div#docTranslation').is(':visible')) {
         translateDoc();
     }
 }


sushain97:
Hm, an even deeper question, why is `translate` even being called? This is a good fix and will be merged but perhaps we can also solve that at the same time.

share-with-me:
> Hm, an even deeper question, why is translate even being called?

Yea! That'd be fixing the root cause too :D 
Um, I saw some files and I realised that in `localization.js`:

`var possibleItems = {'translation': getPairs, 'generation': getGenerators, 'analyzation': getAnalyzers};`

gets called on `$(document).ready...`. In the above snippet, `getPairs` is present in `translator.js`. The code is [here](https://github.com/goavki/apertium-html-tools/blob/master/assets/js/translator.js#L295-L370). Now, the `handlePairs()` gets called which in turn calls the `translate()` method.

Issue #83, opened 2016-12-18
Website translation


sushain97:
This feature has been often requested and used to be present on the old website.

unhammer:
Running in http://gtweb.uit.no/jorgal/ using branch giellatekno of this repo, but several bugs need fixing for this to be considered stable, e.g.

* https://sourceforge.net/p/apertium/tickets/114/ the apertium pipeline needs an overhaul to correctly handle formatting
* https://github.com/goavki/apertium-apy/issues/53 keep pipelines open when translating html pages
* https://github.com/goavki/apertium-apy/issues/52 improve the url cache

sushain97:
some test sites:

![image](https://user-images.githubusercontent.com/4211838/29603772-6ab06fec-87b3-11e7-8bf0-83e771db0cc2.png)

![image](https://user-images.githubusercontent.com/4211838/29603942-ec350262-87b3-11e7-9ea2-a3e5921cb5a0.png)

![image](https://user-images.githubusercontent.com/4211838/29604088-62a9f5a6-87b4-11e7-8544-baa398f9aeb4.png)

![image](https://user-images.githubusercontent.com/4211838/29604111-77a6cc72-87b4-11e7-84f8-3b7c369b4e88.png)

![image](https://user-images.githubusercontent.com/4211838/29604205-c6749212-87b4-11e7-9b9a-f71befdf797f.png)

some also don't work so great so we should follow @jonorthwash's suggestion regarding a "Beta" label.

PR #163, opened 2017-07-05
Installation notification (progress on #83)

sushain97:


sushain97:
@share-with-me I moved here so that we could start the conversation afresh and get rid of the long thread. I made some modifications, primarily some code cleanup. Taking another look, is it necessary to have both a count of requests made and array that contains the request times? It seems like it would be possible to simply take its length.

I don't think the fade is working when I do:

```
installationNotificationShown = false; 
displayInstallationNotification()
```

Also, could you up the size of the alert text a bit? It seems smaller than normal body text. And some padding between the header and text would go a long way towards cleaning it up.

share-with-me:
> is it necessary to have both a count of requests made and array that contains the request times? It seems like it would be possible to simply take its length.

Um, how would you calculate the average of the requests then? I think am not clear with your point. We maintain a queue of request times that helps us to keep the track of average of latest `MAX(numRequests, INSTALLATION_BUFFER)` number of requests!

> I don't think the fade is working when I do:

Um, could you tell where exactly you are executing `displayInstallationNotification()`? I tried in the `callApy()` and on page load, and it does work as expected!

> Also, could you up the size of the alert text a bit? It seems smaller than normal body text. And some padding between the header and text would go a long way towards cleaning it up.

On it :D

sushain97:
> maintain a queue of request times

This queue has a length.

share-with-me:
> This queue has a length.

Ah! makes sense. I misinterpreted it! Will change it now :D

share-with-me:
Um, @sushain97  I checked out this branch and made the necessary changes! I did a `git push origin installationNotification`, and that prompted me to create new PR. Is there a way I can push to this PR?

sushain97:
Not really. Creating a new one is fine.

share-with-me:
Closing here and continuing on #164

Issue #166, opened 2017-07-16
TranslateText() to be called only when detectLanguage() completes


share-with-me:
Currently, on clicking the `identifyLang` button, it calls the `detectlanguage()` function and immediately calls the next method in succession ie, `translateText()`. Ideally, it shoulld get called only when `detectLanguage` completes.

Issue #131, opened 2017-04-12
langdetect: target language button changes but not url-bar, gives "translation not available"


unhammer:
```
<qdb> hello . https://www.apertium.org/ has some bug. i pasted kazakh text and
    target button changed to tatar, but, it said "translation is not
    available". then i see that it was written kaz-spa in address bar.  [20:42]
```

I can reproduce. Go to e.g.
https://www.apertium.org/index.eng.html?dir=eng-spa#translation
then paste in 

*орналасқан Шығыс Азиядағы мемлекет. Астанасы — Сеул. БАҚ өкілдерімен жиі қолданылатын, мемлекеттің бейресми атауы — Оңтүстік Корея.*

and click the "detect language" button. It switches to Kazakh on source and Tatar shows as the only choice on target, but the url field says "kaz-spa" and the translated-text field stays empty.

share-with-me:
Looking at the code of `detectLanguage()`, it seems that only `curSrcLang` is determined in the `success` callback. A `curDstLang` update should also be made according to the `curSrcLang`; which isn't there (and this may be the cause I believe!) Thus, we have source as "kaz" but no where the code in the same callback updates the `destination language buttons` and so forth to have a "tat". I added `autoSelectDstLang()` to the `success` callback with a few modifications and it does give correct output now. However, I am under confident of my solution as I do not understand how I should test it (how do I understand if its working for different languages?)

sushain97:
From @unhammer's bug report, it appears that

1. The destination language is being switched to `tat`.
2. The URL bar is not updated to `kaz-tat`.
3. A new translate request is not fired to update the target language text.

@share-with-me your description of the situation is not consistent with (1) above. Are we experiencing the same bug?

share-with-me:
@sushain97 , um, when I do a `console.log` of the languages as I have shown in the below code, I get the output as in the image after that! The output shows that `curDstLang` is not updated to `tat`! I have made sure that `tat` does not appear in my `recentDstLangs` too. That can be seen in the `console` output. Thus, I believe my observation is not exactly same as the one reported in @unhammer 's report!
@unhammer , could you please correct me if I am unable to find out the snippet wherein the `curDstLang` gets updated as per the detected language?

**Success callback in detectLanguage()**

```
        success: function (data) {
            var possibleLanguages = [];
            for(var lang in data) {
                possibleLanguages.push([lang.indexOf('-') !== -1 ? lang.split('-')[0] : lang, data[lang]]);
[...]

PR #167, opened 2017-07-16
translateText() called only after detectLanguage() completes - fix issue #166

share-with-me:
This PR aims to solve #166 . `translateText` now defers until `detectLanguage` is called. Also, `detectLanguage` is wrapped around `callApy()` method.

### DIFF
--- a/assets/js/translator.js
+++ b/assets/js/translator.js
@@ -169,7 +169,7 @@ if(modeEnabled('translation')) {
         $('#detect').click(function () {
             $('.srcLang').removeClass('active');
             $(this).addClass('active');
-            $.when(detectLanguage()).done(function (x) {
+            $.when(detectLanguage()).done(function () {
                 translateText();
             });
         });
--- a/assets/js/translator.js
+++ b/assets/js/translator.js
@@ -65,18 +65,16 @@ if(modeEnabled('translation')) {
             handleNewCurrentLang(curDstLang, recentDstLangs, 'dstLang');
         });
 
-        $('.srcLang').click(function () {
-            if(this.id !== 'detect') {
-                curSrcLang = $(this).attr('data-code');
-                $('.srcLang').removeClass('active');
-                $(this).addClass('active');
-                populateTranslationList();
-                refreshLangList(true);
-                muteLanguages();
-                localizeInterface();
-                translateText();
-                autoSelectDstLang();
-            }
+        $('.srcLang:not(#detect)').click(function () {
+            curSrcLang = $(this).attr('data-code');
+            $('.srcLang').removeClass('active');
+            $(this).addClass('active');
+            populateTranslationList();
+            refreshLangList(true);
+            muteLanguages();
+            localizeInterface();
+            translateText();
+            autoSelectDstLang();
         });
 
         $('.dstLang').click(function () {
--- a/assets/js/translator.js
+++ b/assets/js/translator.js
@@ -73,7 +73,9 @@ if(modeEnabled('translation')) {
             refreshLangList(true);
             muteLanguages();
             localizeInterface();
-            translateText();
+            if(this.id !== 'detect') {
+                translateText();
+            }
 
             autoSelectDstLang();
         });
@@ -167,8 +169,9 @@ if(modeEnabled('translation')) {
         $('#detect').click(function () {
             $('.srcLang').removeClass('active');
             $(this).addClass('active');
-            detectLanguage();
-            translateText();
+            $.when(detectLanguage()).done(function () {
+                translateText();
+            });
         });
 
         $('.swapLangBtn').click(function () {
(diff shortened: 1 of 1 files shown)


assets/js/translator.js
```diff
@@ -749,50 +753,54 @@ function detectLanguage() {
         textTranslateRequest.abort();
     }
 
-    textTranslateRequest = $.jsonp({
```
sushain97: Looks like we're missing a newline before this function, would you mind adding that in?

assets/js/translator.js
```diff
@@ -749,50 +753,54 @@ function detectLanguage() {
...
-            if(recentSrcLangs.length > TRANSLATION_LIST_BUTTONS) {
-                recentSrcLangs = recentSrcLangs.slice(0, TRANSLATION_LIST_BUTTONS);
-            }
+    return textTranslateRequest;
+}
 
-            curSrcLang = recentSrcLangs[0];
-            $('#srcLangSelect').val(curSrcLang);
-            muteLanguages();
+function detectLanguageSuccessResponse(data) {
```
sushain97: Let's move both of these functions out of global scope and into the `detectLanguage` function? Also let's go with consistent naming here mirroring `success: handleTranslateSuccessResponse,`.

assets/js/translator.js
sushain97: Looks like we're missing a newline before this function, would you mind adding that in?

share-with-me:
> Good work with using a deferred! Minor changes requested that should also hopefully make the diff easier to read...

Thanks @sushain97  :D

assets/js/translator.js
share-with-me: On it.

assets/js/translator.js
sushain97: > There is a newline missing above the detectLanguage() method call though which I have now added.

:+1: this one

share-with-me: :D

assets/js/translator.js
```diff
@@ -73,7 +73,9 @@ if(modeEnabled('translation')) {
             refreshLangList(true);
             muteLanguages();
             localizeInterface();
-            translateText();
+            if(this.id !== 'detect') {
```
sushain97: Can this particular conditional wrap all the function calls to prevent unnecessary work?

assets/js/translator.js
```diff
@@ -167,8 +169,9 @@ if(modeEnabled('translation')) {
         $('#detect').click(function () {
             $('.srcLang').removeClass('active');
             $(this).addClass('active');
-            detectLanguage();
-            translateText();
+            $.when(detectLanguage()).done(function () {
```
sushain97: `$.when(detectLanguage()).done(translateText)` ?

assets/js/translator.js
```diff
@@ -206,8 +209,9 @@ if(modeEnabled('translation')) {
         $('#srcLangSelect').change(function () {
             var selectValue = $(this).val();
             if(selectValue === 'detect') {
-                detectLanguage();
-                translateText();
+                $.when(detectLanguage()).done(function () {
```
sushain97: Same here.

assets/js/translator.js
share-with-me: Yes.

assets/js/translator.js
share-with-me: Yes, this would definitely save up on a lot of redundant work.

share-with-me:
Made the changes @sushain97 !

assets/js/translator.js
```diff
@@ -66,16 +66,17 @@ if(modeEnabled('translation')) {
...
-            $('.srcLang').removeClass('active');
-            $(this).addClass('active');
-            populateTranslationList();
-            refreshLangList(true);
-            muteLanguages();
-            localizeInterface();
-            translateText();
-
-            autoSelectDstLang();
+            if(this.id !== 'detect') {
```
sushain97: If it's literally all of it, then we should just use a different selector, no? `.srcLang:not(#detect)`?

share-with-me:
> .srcLang:not(#detect)

Wow! Implemented :D (I learnt use of more selectors through this :P)

assets/js/translator.js
```diff
@@ -74,7 +74,6 @@ if(modeEnabled('translation')) {
             muteLanguages();
             localizeInterface();
             translateText();
-
             autoSelectDstLang();
```
sushain97: Should L76 and L77 be switched?

share-with-me: Yes makes sense. I did not pay much attention to the inside block :/ . I'd rectify this right away!

share-with-me:
Done!

sushain97:
I tried this out and I don't think it works as intended. The detect button now doesn't run some of the necessary methods such as `autoSelectDstLang`, `muteLanguages`, etc. When I said the following:

> Can this particular conditional wrap all the function calls to prevent unnecessary work?
  
I wasn't certain if it could work. Now I'm fairly certain that it breaks things to never run any of that code on language detection.

share-with-me:
> The detect button now doesn't run some of the necessary methods

Um, aren't those covered in the `success` handler of `detectLanguage` - L787 and L790? There is one method `autoDstSelectLang` which is lacking there which I had mentioned in this issue #131 .I was waiting for this PR to get merged and then create another PR to modify the handler. Should I also fix that issue in this PR? I'll add `autoDstSelectLang` before `muteLanguages` in the handler.

PR #171, opened 2017-07-22
Auto select of destination language for a source language on small screen - Fix #170

### LINKED ISSUE #170: Autoselect of destination language when option is changed on small screen is lacking
Currently, in smaller screens window (zoomed at >= 200%), when a `select` option is changed in `srcLangSelect`, it does not automatically detect a corresponding destination language (which exists in the `pairs` of selected language).

For instance:

In the below image, I choose a language pair: `{eng: glg}`

screenshot from 2017-07-22 13-18-20

On changing the source language from dropdown to Hindi, the destination language still shows only `Galician` 

screenshot from 2017-07-22 13-18-32

It should detect the corresponding language (which is `Urdu`) and set the `dstLangSelect` to that value.

share-with-me:
Fixes #170 . This PR selects a destination language for a selected source language for the views (-xs). This patch also implements the behaviour discussed in this [comment](https://github.com/goavki/apertium-html-tools/issues/170#issuecomment-317164866). - L199. If this behaviour is incorrect, we could simply remove this line.

### DIFF
--- a/assets/js/translator.js
+++ b/assets/js/translator.js
@@ -196,7 +196,6 @@ if(modeEnabled('translation')) {
 
             refreshLangList(true);
             muteLanguages();
-            autoSelectDstLang();
 
             if($('.active > #detectedText')) {
                 $('.srcLang').removeClass('active');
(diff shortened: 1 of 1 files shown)


share-with-me:
Implemented [comment](https://github.com/goavki/apertium-html-tools/issues/170#issuecomment-317265082)

PR #173, opened 2017-07-23
Check for the Analyzer input and return if the input is empty - Fix #172

### LINKED ISSUE #172: 500 Internal server error on empty input for Analyzer mode
If the input for `Analyzer` mode is empty, it gives a `500-Internal Server error` as it expects some input for the mode on the APy side. Ideally, this should be handled on the client side by keeping a check on whether there is some input and in case there is none, the `analyzer request` should be aborted.

screenshot from 2017-07-23 12-49-00

share-with-me:
Fixes #172 . Checks if the input is non-empty. If empty, simply returns.

### DIFF
--- a/assets/js/analyzer.js
+++ b/assets/js/analyzer.js
@@ -143,7 +143,7 @@ function populateSecondaryAnalyzerList() {
 }
 
 function analyze() {
-    if($('#primaryAnalyzerMode').val() === null || $('#morphAnalyzerInput').val() === '') {
+    if($('#primaryAnalyzerMode').val() === null || !$('#morphAnalyzerInput').val()) {
         return;
     }
 
--- a/assets/js/analyzer.js
+++ b/assets/js/analyzer.js
@@ -8,7 +8,7 @@ var currentAnalyzerRequest;
 
 if(modeEnabled('analyzation')) {
     $(document).ready(function () {
-        $('#analyze').click(function () {
+        $('#analysisForm').submit(function () {
             analyze();
             persistChoices('analyzer', true);
         });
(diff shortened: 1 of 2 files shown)


assets/js/analyzer.js
```diff
@@ -143,7 +143,7 @@ function populateSecondaryAnalyzerList() {
 }
 
 function analyze() {
-    if($('#primaryAnalyzerMode').val() === null) {
+    if($('#primaryAnalyzerMode').val() === null || $('#morphAnalyzerInput').val() === '') {
```
sushain97: I think I might prefer a falsey check, something like `!$('#morphAnalyzerInput').val()`? Also, does this problem also exist with generation? If so, would appreciate an analogous fix in this PR.

share-with-me: > Also, does this problem also exist with generation?

For generation, the request does pass for empty input successfully and gives an empty result. However, the text input box still expects a text with `Please fill out this field` warning. Um, this seems bit contradictory. Either it should not show that warning or the `generate()` should not let be run for empty inputs. What could be semantically right behaviour? (Or is it okay this way?) @sushain97 ?

screenshot from 2017-07-24 18-44-16

sushain97: I agree. Idea: make the analyzer input required and before executing generate or analyze, do a form validity check.

share-with-me: Um, looking at the code, both analyzer and generate modes have `required` as a part of HTML. Should I implement a similar check as I have done for `analyzer` mode in the `generate` mode?

sushain97: The best solution is triggering the browser's validation AND not actually making the request.

share-with-me:
@sushain97 , validation is now done at browser :) . If the input in text boxes is empty, the request won't be even made.

sushain97:
Great work finding and fixing these smaller issues!

share-with-me:
> Great work finding and fixing these smaller issues! Really helps improve quality.

Thank you so much @sushain97 :D. I have learnt to focus on quality from you!

PR #175, opened 2017-07-25
Disable detectLang on translateDoc interface, enable otherwise

### LINKED ISSUE #174: Disable 'detectLanguage' when on 'docTranslation' interface?
When on `docTranslation` interface, on clicking `detect`, it takes up the contents of `originalText` and does the detect. But, one would expect the functionality to detect the language of the contents inside the uploaded doc thereby making it misleading. **Does the `detectLanguage` functionality work for docs too?**
(I could only see the functionality for `translateText` interface).

If no, would it make sense to disable the `detect` button while on `docTranslation` interface and then enable it again while switching to `translateText` interface?

share-with-me:
Fix #174

### DIFF
--- a/assets/js/translator.js
+++ b/assets/js/translator.js
@@ -223,8 +223,7 @@ if(modeEnabled('translation')) {
                 $('#fileInput').show();
                 $('div#fileName').hide();
                 $('div#docTranslation').fadeIn('fast');
-                $('#detect').prop('disabled', true);
-                $('#srcLangSelect').children('option[value=' + 'detect' + ']').attr('disabled', true);
+                $('#detect, #srcLangSelect option[value=detect]').prop('disabled', true);
             });
             pairs = originalPairs;
             populateTranslationList();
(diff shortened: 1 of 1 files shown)


assets/js/translator.js
```diff
@@ -223,6 +223,8 @@ if(modeEnabled('translation')) {
                 $('#fileInput').show();
                 $('div#fileName').hide();
                 $('div#docTranslation').fadeIn('fast');
+                $('#detect').prop('disabled', true);
+                $('#srcLangSelect').children('option[value=' + 'detect' + ']').attr('disabled', true);
```
sushain97: Why are you using concatenation here?

sushain97: I would suggest actually combining these into a "single" selector with a comma.

share-with-me: I followed the style as used across the code, but yeah, we are looking only for a constant `string`, I'd change it now!

share-with-me: Awesome, makes sense

share-with-me: Um @sushain97 , are you suggesting something like following? (using a jquery inside another selector)?  The `option` value does not have an ID. Thus I'll need to select `detect` from the `select` dropdown and disable the option. Referred [this](https://stackoverflow.com/questions/10525651/jquery-selector-element-inside-other)

`$('#detect', $('#srcLangSelect').children('option[value=detect]')).attr('disabled', true);`

sushain97: Why are you using children?

share-with-me:
Combined in selector. Sorry for poor glitches on this one :) . I was using the `select option` through jquery API for first time. Have read the API usages now.

Thanks!

PR #179, opened 2017-08-01
Analyze and Generate buttons aligned

### LINKED ISSUE #177: Alignment of buttons on interfaces
The buttons for various functionalities should be aligned with the text box and the language dropdown above it. For eg, in the below image, the left edge of `Analyse` button should be aligned as above elements.

screenshot from 2017-07-31 19-47-06

share-with-me:
Fix #177


sushain97:
I think that the APY sandbox needs the same fix?

share-with-me:
> I think that the APY sandbox needs the same fix?

Yes. I have never used that mode interface and it completely went off the radar :P .Would change and push it now.

PR #180, opened 2017-08-02
Markunknown sent with translateDoc interface

share-with-me:
Fix #160

### DIFF
--- a/assets/js/translator.js
+++ b/assets/js/translator.js
@@ -218,12 +218,13 @@ if(modeEnabled('translation')) {
         });
 
         $('button#translateDoc').click(function () {
-            $('div#translateText').fadeOut('fast', function () {
-                $('#fileInput').show();
-                $('div#fileName').hide();
-                $('div#docTranslation').fadeIn('fast');
-                $('#detect, #srcLangSelect option[value=detect]').prop('disabled', true);
-            });
+            $('#translateText > *:not(#translateOptionsContainer), #translateOptions > *:not(#markUnknownContainer')
+                .fadeOut('fast', function () {
+                    $('#fileInput').show();
+                    $('div#fileName').hide();
+                    $('div#docTranslation').fadeIn('fast');
+                    $('#detect, #srcLangSelect option[value=detect]').prop('disabled', true);
+                });
             pairs = originalPairs;
             populateTranslationList();
         });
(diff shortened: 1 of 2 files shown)


assets/js/translator.js
```diff
@@ -218,12 +218,13 @@ if(modeEnabled('translation')) {
         });
 
         $('button#translateDoc').click(function () {
-            $('div#translateText').fadeOut('fast', function () {
-                $('#fileInput').show();
-                $('div#fileName').hide();
-                $('div#docTranslation').fadeIn('fast');
-                $('#detect, #srcLangSelect option[value=detect]').prop('disabled', true);
-            });
+            $('#translateText > *:not(#translateOptionsContainer), #translateOptions > *:not(#markUnknownContainer')
```
sushain97: I feel like this selector is missing a closing paren.

share-with-me: Felt like I closed it with the paren outside. Added :D

sushain97:
image

This ends up looking really strange. The option should be in a similar position to where it is usually located.

share-with-me:
@sushain97 , Um, would it make sense to simply add a CSS of `margin-top` to the checkbox? Tried quite a bit messing around with `visibility` and `display` attributes, but the way the divs are arranged, they really are taking up weird places when I use the above attributes on them!

sushain97:
> @sushain97 , Um, would it make sense to simply add a CSS of margin-top to the checkbox?

Then it'll be too low when a file gets translated?

share-with-me:
@sushain97 , position set as required. Please have a look.

screenshot from 2017-08-10 20-41-42

sushain97:
Why is it all the way at the bottom and not where it used to be?

share-with-me:
I am preserving its position wrt to the translation interface. Thus, I am adding a top margin equivalent to the height of `translatedText` div. It looks very much at the bottom because of the relatively smaller height of `docTranslation` div and the absence of `Instant Translation` checkbox. What do you suggest its ideal height should be? I felt keeping it at the height where `docTranslation` ends would make more sense. Should I change it to that?

sushain97:
> I felt keeping it at the height where docTranslation ends would make more sense. Should I change it to that?
> 

Yes. This is probably better.

PR #187, opened 2017-08-15
apyRequestTimeout cleared on request abort - Fix #185

### LINKED ISSUE #185: Clear apyRequestTimeout in the event of request abort
In case of successive calls of the same AJAX method, the prior ones are aborted and only the last one gets executed. However, the prior requests do start the `apyRequestTimeout` timer in the `callApy` method and when aborted, cause this `apyRequestTimeout` variable to display the installation notification.

For instance, when `detectLanguage` is called successively (See console) , the first request is aborted. However, as the `apyRequestTimeout` variable which is set by this request is not cleared, it displays the notification erroneously.

screenshot from 2017-08-15 00-45-03

share-with-me:
Fix #185 .

### DIFF
--- a/assets/js/analyzer.js
+++ b/assets/js/analyzer.js
@@ -3,7 +3,7 @@ var currentAnalyzerRequest;
 
 /* exported getAnalyzers */
 /* global config, modeEnabled, persistChoices, restoreChoices, localizeInterface, readCache, ajaxSend, ajaxComplete,
-    cache, getLangByCode, filterLangList, allowedLang, sendEvent, callApy */
+    cache, getLangByCode, filterLangList, allowedLang, sendEvent, callApy, apyRequestTimeout */
 /* global ENTER_KEY_CODE */
 
 if(modeEnabled('analyzation')) {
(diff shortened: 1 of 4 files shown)


sushain97:
Is there not an abort callback or something equivalent?

share-with-me:
You mean a function gets called in the event `abort()` is made as callback? And that function clears the `apyRequestTimeout` ?

sushain97:
Yes. I'm not a fan of the massive repetition here.

share-with-me:
On it.

share-with-me:
@sushain97 , would it make sense if I handle this in the `error` callback for these requests? For instance, in all the error callbacks, I would add this:

```
if(textStatus === 'abort') {
   clearTimeout(apyRequestTimeout);
}
```

This would make sense from the point of view of error handling but I still can't understand if this would help in reducing repetition. A lot of Google search yielded that it could be handled in the error callback of the requests.

sushain97:
That approach makes sense to me. If those two lines only exist inside `callApy`, we’ve reduced repetition.

share-with-me:
@sushain97 , Um, I tried to check for different callbacks that get called in case the request is aborted. If the request is sent and then the request is aborted, the `success` handler is simply ignored. On the other hand, if `abort` is made before request is sent, `error` callback is called wherein the above code snippet could be used to achieve the functionality. I tried `always` and `fail` params with the request. However, they do not get executed in the event of abort. I am finding a bit difficult to obtain an AJAX handler parameter that is always called in the event request is aborted. Could you help a bit in this regard?

PR #189, opened 2017-08-18
Enter does not call analyze, generate, sanbox requests on empty input - Fix #188

### LINKED ISSUE #188: Enter button executes analyze, generate, sandbox methods even on empty input
screenshot from 2017-08-18 00-43-12

share-with-me:
Fix #188

### DIFF
--- a/assets/js/analyzer.js
+++ b/assets/js/analyzer.js
@@ -143,9 +143,9 @@ function populateSecondaryAnalyzerList() {
 }
 
 function analyze() {
-    var morphAnalyzerInput = $('#morphAnalyzerInput').val();
+    var input = $('#morphAnalyzerInput').val();
 
-    if($('#primaryAnalyzerMode').val() === null || morphAnalyzerInput.trim() === '') {
+    if($('#primaryAnalyzerMode').val() === null || input.trim() === '') {
         return;
     }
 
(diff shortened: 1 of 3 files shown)


assets/js/analyzer.js
```diff
@@ -26,7 +26,9 @@ if(modeEnabled('analyzation')) {
         $('#morphAnalyzerInput').keydown(function (e) {
             if(e.keyCode === ENTER_KEY_CODE && !e.shiftKey) {
                 e.preventDefault();
-                analyze();
+                if($('#morphAnalyzerInput').val().trim() !== '') {
```
sushain97: I think this might actually be cleaner if the check is performed in `analyze`. There's already an early return there. `$('#morphAnalyzerInput').val()` can also probably be assigned to a variable since it gets used later. Similar comment for the rest of these changes.

share-with-me: Ohh. Awesome. I just now figured that submitting the forms with nothing but spaces by clicking on button also gives the error. Shifting this validation there would fix that error too! 

screenshot from 2017-08-18 09-41-00

I submit the form with spaces and then it gives error.

assets/js/analyzer.js
```diff
@@ -143,7 +143,9 @@ function populateSecondaryAnalyzerList() {
 }
 
 function analyze() {
-    if($('#primaryAnalyzerMode').val() === null) {
+    var morphAnalyzerInput = $('#morphAnalyzerInput').val();
```
sushain97: Can we just makes these all `input`?

PR #202, opened 2017-08-23
Website translation

sushain97:


sushain97:
- [ ] does not accept URLs without http/https directly in input box
- [ ] does not actually translate page on refresh after translating a page with hash containing URL
- [ ] this ugliness:

image

@share-with-me this is why we don't hardcode values :) They break on resize.

sushain97:
@share-with-me I've merged the APy part of this since it seems to work. I think you may want to reduce the timeout on the web request to the external site since currently it always triggers the installation notification.

More importantly, this PR is mergeable after the issues above are fixed. Took a couple hours today to clean up a bunch of this code!

share-with-me:
> More importantly, this PR is mergeable after the issues above are fixed. Took a couple hours today to clean up a bunch of this code!

Thank you so much @sushain97 . I really appreciate the time you put in this. It means a lot to me. I do agree that this PR involves a lot of code changes and I shall make sure that everything works  :D.

share-with-me:
>  I think you may want to reduce the timeout on the web request to the external site since currently it always triggers the installation notification.

Um, do you mean increase the timeout? So that, the webpage translation request has ample time to execute? (and thereby not trigger the notification)?

sushain97:
> Um, do you mean increase the timeout? So that, the webpage translation request has ample time to execute? (and thereby not trigger the notification)?

No. Fetching an external website has a 20 second timeout in APy right now. That's probably unnecessarily long.

share-with-me:
> No, that's the opposite of what I see as the right solution. Fetching an external website has a 20 second timeout in APy right now. That's probably unnecessarily long.

Ohh! I was referring timeout as the notification timeout for the frontend. I interpreted in a wrong way! Um, so I tried making the connection timeout value to 5 second in APy. It did give the notification on the frontend as well as raised the error of Connection timeout.

sushain97:
> Um, so I tried making the connection timeout value to 5 second in APy. It did give the notification on the frontend as well as raised the error of Connection timeout.

That's alright. We can file an issue to increase the Html-tools APy request notification timeout for this particular endpoint.

sushain97:
> This is also the reason why the spurious requests are getting called when Cancel is clicked.

I think that's because the cancel button is triggering the form submit. It should probably be changed to `type="button"` to disable that behavior.

> thereby it will result in Failed translation requests.

It should never do that. Those methods should exit out early if there's nothing to send. I know document translation does.

share-with-me:
>  It should probably be changed to type="button" to disable that behavior.

This is strange. All of the buttons in that form have `type=button` and yet, the form is submitted.

share-with-me:
Everyting seems to be working like the original behaviour when I remove this snippet:

```
        $('#translationForm').submit(function () {
            translate();
            persistChoices('translator', true);
        });
```

sushain97:
> Everyting seems to be working like the original behaviour when I remove this snippet:

You're saying that all the following work:

1) Instant text translation
2) Text translation clicking "Translate" button
3) Document translation clicking "Translate" button
4) Document translation dragging document onto page
5) Instant website translation by typing name
6) Website translation clicking "Translate" button
7) Website translation typing URL in input#webpage and hitting `Enter` key

If all 7 workflows work, then remove it.

share-with-me:
> Website translation typing URL in input#webpage and hitting Enter key

The above step fails and it gives `Please enter a URL` browser validation. And that is because of `input type='url'` I believe. The rest of 6 steps did work.

sushain97:
> The above step fails and it gives Please enter a URL browser validation. And that is because of input type='url' I believe. 

Then use a real URL...

share-with-me:
Nope. It does not do anyting on inputting a URL and clicking on Enter. I believe that functionality should be coded? (As in it is dependent on the Javascript that we would write?)

sushain97:
> I believe that functionality should be coded? (As in it is dependent on the Javascript that we would write?)

No. It should pass through the form submit handler which you removed.

share-with-me:
> No. It should pass through the form submit handler which you removed.

Um, the scripts in `analyzer.js` and `generator.js` do make use of forms and submit. However, we do have explicitly coded the functionality in the event ENTER key is pressed as follows (why did it not pass the values through the form submit handler in this case?):

```
            if(e.keyCode === ENTER_KEY_CODE && !e.shiftKey) {
                e.preventDefault();
                analyze();
            }
        });
```

I was planning to do something like above.

sushain97:
> why did it not pass the values through the form submit handler in this case?

Because those inputs are `textarea`'s. This is an `input` (single line) which fires the submit event on enter, `textarea`s do not.

> I was planning to do something like above.

Yes. From my previous comment:

> Website translation typing URL in input#webpage and hitting Enter key (this shouldn't require a special handler that checks for enter keys).

share-with-me:
Ohh! Thanks @sushain97 . I'll try to resolve this now.

share-with-me:
Um, I have spent some time to understand why the form is getting submitted and am unable to figure out why the form gets submitted even when the button is explicitly made of type `button` :/ . I believe rather than making validations that form shouldn't get submitted when the inputs are empty as well ensuring that `Cancel` shouldn't submit the form, would it make sense to have a code that lets the ENTER work for url though javascript (I have written that) ? We could tackle about the form submission issue in the issue to be opened for following?: 

` does not accept URLs without http/https directly in input box [DON'T DO THIS NOW -- CREATE AN ISSUE AFTER THIS MERGES]`

sushain97:
> Could you help out a bit here

Unfortunately, I don't have any time to help with this today. I would use the JavaScript debugger provided with Chrome.

> I believe rather than making validations that form shouldn't get submitted when the inputs are empty

We already do this. If you want to remove all that validation, we should add `required` to the inputs and work primarily through the `submit` event so that the browser's validation catches the problem and shows the user a localized error rather than inserting early returns into our code.

> ensuring that Cancel shouldn't submit the form

I consider it a bug that this happens at all even if we don't actually handle the event.

> would it make sense to have a code that lets the ENTER work for url though javascript

Not really, that's why I removed the code that behaved that way. The application should use as little JS as possible when it can rely on the browser's default behavior instead.

If you can't find a way to do it the right way, go ahead and replace the submit handler with the keypress/Enter handler. However, the other issues I mentioned **need** to be fixed the proper way since we shouldn't be introducing a bunch of hacky solutions.

share-with-me:
Yes. I shall try looking for its solution and if I find nothing, I shall temporarily replace the submit handler with the keypress/Enter handler. The other issues which are fixed:

-  does not actually translate page on refresh after translating a page with hash containing URL

If I replace the submit handler with Enter handler, the following issue gets fixed.

-  spurious "translation not available" occurs on document translation cancel, also failed translation request.

The issue that is then pending is that of:

- ugliness

share-with-me:
Continuing in #204 . I am unsure whether to close this.

PR #209, opened 2017-10-15
Mark multi-step destinations

### LINKED ISSUE #120: mark destinations through multi-step translation somehow
Currently when multi-step is enabled, only the option says "experimental" and not the translation destination languages.  We should have some way of indicating that these are experimental "pairs".

shardulc:
Marks multi-step destinations with a '+' sign, explanation string added only for English. (Attempt to fix #120)

### DIFF
--- a/assets/css/style.css
+++ b/assets/css/style.css
@@ -221,6 +221,38 @@ html[dir='rtl'] .languageName {
     cursor: pointer;
 }
 
+.languageNameTooltip {
+    visibility: hidden;
+    width: auto;
+    background-color: #555;
+    color: #fff;
+    text-align: center;
+    border-radius: 6px;
+    padding: 2px 5px;
+    position: absolute;
+    z-index: 1;
+    top: 125%;
+    left: 0%;
+    opacity: 0;
+    transition: opacity .3s;
+}
+
+.languageName .languageNameTooltip::after {
+   content: "";
+   position: absolute;
+   bottom: 100%;
+   left: 50%;
+   margin-left: -5px;
+   border-width: 5px;
+   border-style: solid;
+   border-color: transparent transparent black transparent;
+}
+
+.languageName:hover:not(.text-muted) .languageNameTooltip {
+    visibility: visible;
+    opacity: 1;
+}
+
 html[dir='ltr'] #srcLanguages {
     min-width: 500px;
     right: auto;
--- a/assets/css/style.css
+++ b/assets/css/style.css
@@ -436,7 +436,6 @@ html[dir='rtl'] .pull-left:not(#rtlChecker) { /* sass-lint:disable-line class-na
 }
 
 .langs {
-    -khtml-user-select: none;
     -moz-user-select: none;
     -ms-user-select: none;
     -webkit-touch-callout: none;
--- a/assets/css/style.css
+++ b/assets/css/style.css
@@ -221,36 +221,12 @@ html[dir='rtl'] .languageName {
     cursor: pointer;
 }
 
-.languageNameTooltip {
-    background-color: #555;
-    border-radius: 6px;
-    color: #fff;
-    left: 0%;
-    opacity: 0;
-    padding: 2px 5px;
-    position: absolute;
-    text-align: center;
-    top: 125%;
-    transition: opacity .4s;
-    visibility: hidden;
-    width: auto;
-    z-index: 1;
+#dstLanguages .tooltip.bottom {
+    padding-top: 4px;
 }
 
-.languageName .languageNameTooltip::after {
-    border-color: transparent transparent #000;
-    border-style: solid;
-    border-width: 5px;
-    bottom: 100%;
-    content: '';
-    left: 50%;
-    margin-left: -5px;
-    position: absolute;
-}
-
-.languageName:hover:not(.text-muted) .languageNameTooltip {
-    opacity: 1;
-    visibility: visible;
+#dstLanguages .tooltip.bottom > .tooltip.arrow {
+    border-bottom: 6px solid #000;
 }
 
 html[dir='ltr'] #srcLanguages {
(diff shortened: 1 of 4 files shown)


sushain97:
Mind throwing up some screenshots to make this stuff easier to make suggestions on? Running an APy instance that actually supports multi-step translation with enough pairs to make the graph meaningful can be a little annoying ;)

shardulc:
Here's a screenshot with multi-step translation between all pairs supported on the Apertium APy instance.

image

sushain97:
Hm, okay. I think it makes sense for the non-multi-step target languages to come first. Thoughts? @jonorthwash 

Also, I'm not super sold on the "+". Maybe some sort of a symbol? Arrow with two marks on it? Something that looks like an angle symbol? Unsure. The plus is a bit confusing though.

shardulc:
I agree that the direct targets should come first. For the symbol, do you think there could be a non-symbol way of representing it (a bad example is coloring the text)?

sushain97:
Unfortunately, nothing great comes to mind immediately. What might be really cool is showing the path that the translation takes when you hover over the symbol in a tooltip.

shardulc:
Updated screenshot, still using '+'.
image

jonorthwash:
I also don't like the '+'.  It has connotations of "better" or "more advanced", which in this case is false.  I'd prefer '(-_n_)' (in red even), where _n_ is the number of steps it has to go through.

shardulc:
Updated screenshots:

screenshot0

screenshot1

What should the checkbox text say instead of "experimental, marked with +"? Talking about weights isn't for everyone?

shardulc:
To get the weights from APy, it currently makes the request `/listPairs?q=pairs&weights=yes`, which is what the APy endpoint should be in the future if weights are supported.

sushain97:
It's looking much better. Some suggestions:

- We need fewer colors. Even if had a key, the current number is a bit too high. Maybe we can make the light blue edges the same gray as the arrows? 
- It doesn't seem that we've highlighted the edges in valid paths as a different color than the current selected path?
- I think we can omit the (-3) on the valid paths list. It's somewhat obvious. Maybe also sort by weight, i.e. the shortest ones first now.
- The tooltip in the destination list doesn't have the little caret that indicates where it came from? It seems randomly placed.

jonorthwash:
I agree with everything @sushain97 said but one thing:

I like the indications of how long the path is, like (-3).  I also think the valid paths should be sorted in order of shortest to longest (and within that, perhaps by some quality of translation measure, like number of stems or something, which we can worry about later).

shardulc:
"We need fewer colors": I agree. The main function of the blue arrows is
that if they aren't there, then often nothing obvious will happen when the
user selects a node, which might be confusing.

"valid paths" vs. "current selected path": Sorry, I don't understand what
you're talking about. Is it that the path that is currently being used
should be shown permanently on the graph?

I disagree with not showing the "path length" because it isn't actually a
path length; if/when in the future we have actual weights to the pairs, the
total weight will be shown, which might be contrary to the actual length
ordering.

"the little caret": ? I don't think I know what this is o.O

sushain97:
> "the little caret": ? I don't think I know what this is o.O

image

Note the small triangle shape (caret) at the bottom of the black tooltip.

> "valid paths" vs. "current selected path": Sorry, I don't understand what
you're talking about. Is it that the path that is currently being used
should be shown permanently on the graph?

I think I've forgotten how this interface actually interacts with the user. Were you able to set up Circle on your fork so that I can test it out?

sushain97:
I changed some CircleCI settings. Try pushing to this branch again and it might build now

shardulc:
@sushain97 [This build][1] fixes the caret issues, and handles a lot of other minor broken things too. I think whether or not to display weights is pretty much resolved. I also think the yellow endpoints and green valid path arrows are self-explanatory, but that might just be me, and the blue arrows are probably not the best idea but we need something else to indicate to the user that selecting a node actually does something.

 [1]: https://12-78226301-gh.circle-artifacts.com/0/home/circleci/project/build/index.html

sushain97:
> the blue arrows are probably not the best idea but we need something else to indicate to the user that selecting a node actually does something.

What do you think about messing with the opacity of the edges that exist in the graph but do not belong to clicked nodes instead? i.e. they would be 50% transparent until clicked? Seems to be more natural/semantic.

sushain97:
@shardulc could you link the latest version? Also I'm thinking that we need some brief explanatory text in the modal to explain that the workflow is 1) click nodes you want to include in the graph 2) click one of the valid paths. Otherwise, it's clear to us since we've used it multiple times but relatively unclear otherwise.

shardulc:
@sushain97 [Here](https://13-78226301-gh.circle-artifacts.com/0/home/circleci/project/build/index.html) is a link to the latest build. Proposal for explanatory text: "Build a path for translation in the graph below by clicking on the desired intermediate languages. When a complete path is built, select it from the 'Valid Paths' list." (Would this be too much effort for localization?) I guess this should be placed above the graph.

shardulc:
@sushain97 Thoughts about this? (There's not much left to do on the technical side so I'm trying to get this done before master accumulates other changes :P)

sushain97:
I'll take a careful look today. Sorry for the delay!

On Nov 16, 2017 6:40 PM, "Shardul Chiplunkar" <notifications@github.com>
wrote:

@sushain97 <https://github.com/sushain97> Thoughts about this? (There's not
much left to do on the technical side so I'm trying to get this done before
master accumulates other changes :P)

—

<https://github.com/goavki/apertium-html-tools/pull/209#issuecomment-345098660>,
or mute the thread

IlnarSelimcan:
This feature is pretty cool and useful! Why hasn't it been enabled on beta.apertium.org?

sushain97:
It isn't in `master`.

PR #213, opened 2017-11-20
Final touches to translation chain graph

shardulc:
@sushain97 I've resolved most of your comments on #211 and I'm recreating those that I haven't.


assets/css/style.css
```diff
@@ -406,7 +406,7 @@ html[dir='rtl'] .pull-left:not(#rtlChecker) { /* sass-lint:disable-line class-na
     float: right !important; /* sass-lint:disable-line no-important */
 }
 
-/* D3 path chooser stuff */
+/* translation path chooser stuff */
```
shardulc: The "choose" terminology is used throughout my code: "chooser", "chosen", etc. If there is some specific reason you prefer "select" over "choose", I'll go through everything and change it.

assets/js/translator.js
```diff
@@ -446,22 +449,26 @@ function initChainGraph() {
...
 function boundary(dist, max) {
-    if(dist < nodeSize) return nodeSize;
-    if(dist < max - nodeSize) return dist;
-    return max - nodeSize;
+    if(dist < CHAIN_NODE_SIZE) return CHAIN_NODE_SIZE;
+    if(dist < max - CHAIN_NODE_SIZE) return dist;
+    return max - CHAIN_NODE_SIZE;
 }
 
 function getAllPaths(src, trgt, curPath, seens) {
```
shardulc: All the `trgt`s (possibly even outside my changes) should be changed into `dst`s, right? I'll leave this as a final change after all everything else is settled because the more important changes will be swamped out.

shardulc:
Current proposals for the checkbox text:
 1. "[Multi-step translation][1] (experimental)"
 2. "[Multi-step translation][1] (experimental, marked with badness (-n))"
 3. "[Multi-step translation][1] (experimental, marked with [weight (-n)][link to wiki page about weights])"

Current proposals for explanatory text just above the graph:
 1. "Build a translation path in the graph below by clicking on the desired intermediate languages. When valid path(s) are formed by the languages you click, select one from the list below."
 2. "Build a translation path in the graph below by clicking on the desired intermediate languages. When one or more valid paths are formed, select the desired one from the 'Valid Paths' below."

@sushain97, @jonorthwash Thoughts?

 [1]: https://en.wikipedia.org/wiki/Pivot_language

jonorthwash:
> Current proposals for the checkbox text:

How about
"[Multi-step translation][1] ([experimental!](http://wiki.apertium.org/wiki/Weights))"

Maybe with "experimental" in a different colour / with different bg colour?

> Current proposals for explanatory text just above the graph:

Either is fine.  Maybe the second one is a tad clearer?

 [1]: https://en.wikipedia.org/wiki/Pivot_language

sushain97:
> Maybe with "experimental" in a different colour / with different bg colour?

I'm not sure even this is needed. It's a pretty short label.

> Either is fine. Maybe the second one is a tad clearer?

I think the "the desired" is extraneous from the second one. Otherwise, it's fine.

assets/js/translator.js
sushain97: > All the trgts (possibly even outside my changes) should be changed into dsts, right?

Yep.

> I'll leave this as a final change after all everything else is settled because the more important changes will be swamped out.

Up to you! I review the entire diff again either way.

assets/css/style.css
sushain97: My primary concern is that the word "chooser" is somewhat awkward in (at least American) English.

e.g.

image

image

assets/js/translator.js
shardulc: I realized APy returns objects with the "target" terminology, not "destination". It makes more sense to change html-tools to match APy, right?

shardulc:
[Latest build.](https://19-78226301-gh.circle-artifacts.com/0/home/circleci/project/build/index.html) The only thing remaining is 'target' vs. 'destination' as in comments above.

assets/js/translator.js
sushain97: We could but I'm somewhat opposed to this because it involves changing a lot of existing code that is relatively brittle, especially without tests.

sushain97:
couple notes:

- "experimental" shouldn't be linked until we have an actual page to link to. I suggest leaving it as before for now and we can relatively easily add links later. I also don't think the exclamation point adds much :)
- use `.lead` on the "Build a translation path" text
- the "-1" still looks too much like "minus 1". I suggest just omitting the symbol entirely; it won't mean much as a weight anyway.
- checking multi-step translation for the first time results in the change interface being blank
- selecting nodes makes the entire interface jump very unnaturally. maybe col-md-3 for the valid paths and col-md-9 for the actual graph will fix that
- there's no indication of what the current selected path is if you change it. maybe this can be fixed by changing the button to "Change translation path ([current path])"

assets/js/translator.js
```diff
@@ -538,13 +538,13 @@ function displayPaths(paths) {
     graph.nodes = nodes;
     graph.links = [];
     Object.keys(arrows).forEach(function (src) {
-        arrows[src].forEach(function (trgt) {
-            if(arrows[trgt] && arrows[trgt].indexOf(src) !== -1) {
-                graph.links.push({'source': src, 'target': trgt, 'right': true, 'left': true});
-                arrows[trgt].splice(arrows[trgt].indexOf(src), 1);
+        arrows[src].forEach(function (dst) {
+            if(arrows[dst] && arrows[dst].indexOf(src) !== -1) {
+                graph.links.push({'source': src, 'target': dst, 'right': true, 'left': true});
```
shardulc: The `target` here (and in other similar contexts) is d3.js syntax for specifying graph links.

shardulc:
[Latest build.](https://23-78226301-gh.circle-artifacts.com/0/home/circleci/project/build/index.html) Should I start working on the wiki page about weights, or is that better made into a GCI task? :P

PR #216, opened 2017-12-02
Make constants easier configurable

### LINKED ISSUE #158: More dynamically configurable options
We have lots of constants inside the application but most of them aren't accessible through `config.conf`. We might want to figure out a more dynamic approach to this rather than adding a line to `read-conf.py` for each new constant.

Androbin:
Fixes #158 

The script `tools/read-conf.py` now lists and describes all constants in `config.conf` as well as the new `custom.conf` which are also extracted from various JS files.
This approach allows for more dynamically configurable options and is easily adaptable.


shardulc:
@Androbin I think what @sushain97 meant in #158 was that configuration options like all the constants [here](https://github.com/goavki/apertium-html-tools/blob/master/assets/js/translator.js#L10) should be somehow dynamically put into a configuration file with the appropriate sections. That way, all constants which can be customized can be customized through a single file, and constants can be easily accessed through any part of html-tools. (@sushain97 Correct me if I'm wrong!)

Your changes make it so that the _existing_ options are easy to change and add manually, and the code is definitely more maintainable, but they do not solve the problem above. Also, they [break the interface entirely](https://52-28576409-gh.circle-artifacts.com/0/home/circleci/project/build/index.eng.html?dir=ukr-rus#translation) so you need to test your changes.

(P.S. Don't squash your commits on the PR, we'll take care of that when we merge! :smiley:)

sushain97:
Yes, *exactly* what @shardulc said :) Read my mind.

sushain97:
For tasks like this one where the issue doesn't exactly spell out a solution, providing a straw man proposal before making the changes is generally a good idea.

Androbin:
@shardulc I have verified that all options get set correctly. What have I missed?
@sushain97 What about this? I could port the proposed python code to javascript and import it from the files that need it.

sushain97:
> @shardulc I have verified that all options get set correctly. What have I missed?

Check whether there are any JS errors and if the built JS looks the same as it used to.

> @sushain97 What about this? I could port the proposed python code to javascript and import it from the files that need it.

Adding a dependency on Node.js or another JS runner isn't acceptable unfortunately.

Androbin:
> Check whether there are any JS errors and if the built JS looks the same as it used to.

But I have not touched any JavaScript code yet.
> Adding a dependency on Node.js or another JS runner isn't acceptable unfortunately.

Don't get me wrong, @shardulc said 
> [...] constants can be easily accessed through any part of html-tools.

which I read as "one central config file" (conf-fields.txt) and language-specific code to read from that file.

sushain97:
> But I have not touched any JavaScript code yet.

The Python script is used in the generation of JS. Therefore, your changes can easily result in changes in the JS.

> language-specific code to read from that file

This might work provided that the config file is still in a standard format (e.g. YML or INI).

Androbin:
I have just realized that a port to JS might not be needed. If I got this right, [`tools/read-conf.py > print_js`](https://github.com/goavki/apertium-html-tools/pull/216/files#diff-bbd3716c3cae20b9a539236beb8f406eR88) injects a [config global](https://github.com/goavki/apertium-html-tools/blob/master/assets/js/translator.js#L24) into JS.
If I move [all the constants](https://github.com/goavki/apertium-html-tools/blob/master/assets/js/translator.js#L10) into the [config file](https://github.com/Androbin/apertium-html-tools/blob/537e9a648e37df1de07ddbe104616a5ae3051050/tools/conf-fields.txt), then JS should be able to access them.

sushain97:
> If I move all the constants into the config file, then JS should be able to access them.

I don't see this as ideal. There are lots of these and they change often. e.g. if a code refactor renames one of them, it should not require everyone changing their configuration file. I'm fine with everyone needing to change their configuration file if they want their custom value to still work but not requiring them to for the interface to function with defaults. Either way, all the constants shouldn't all be declared in multiple places. Perhaps the ones that are overridable can have a comment in the JS indicating as much.

shardulc:
I think a good solution would be to have two config files: one for 'options' like enabling modes etc., which we already have, and one for 'customizations' of all the constants that are currently within the JS files, which is what you would need to create. Both these files would be read by the same config parser and injected as the same global config.

Androbin:
@shardulc Where would you prefer the config files to be located?

shardulc:
`config.conf` (or rather `config.conf.example`) already exists in the top directory, so I'd say you add the other config file in the top directory too. Try to name it something which clearly marks it as being different from `config.conf`.

Androbin:
Shall I de-duplicate [`tools/conf-fields.txt`](https://github.com/Androbin/apertium-html-tools/blob/537e9a648e37df1de07ddbe104616a5ae3051050/tools/conf-fields.txt) and `config.conf` then?

shardulc:
Yes. Personally I think the type of each constant can be in a dict in `tools/read-conf.py`, although @sushain97 might have a better idea.

Androbin:
@shardulc But the dict could only hold 'options' but not 'customizations', right? What about comments behind each entry?

shardulc:
Why not both? And what comments? (By the way, I would prefer to have this discussion on [IRC](http://wiki.apertium.org/wiki/IRC).)

Androbin:
I have not yet moved the new constants into `config-custom.conf` but other than that, it should work now.

Androbin:
@sushain97 Do you approve the proposal so far?

shardulc:
@Androbin I was thinking that we would not have `config-custom.conf.example` at all, instead directly putting `config-custom.conf` under version control (unlike `config.conf`). Also, the build is still not successful, see the second-last step [here](https://circleci.com/gh/goavki/apertium-html-tools/55#artifacts/containers/0) (I don't know why it's showing up as successful...).

Androbin:
@shardulc Alright, I have figured it out. There is a lower/upper case issue with `configparser`. Gimme a moment to fix this.

Androbin:
Also, some values were not set, but should be set to `None` which didn't work by default.

shardulc:
@Androbin A successful build will produce `index.html` as an artifact in CircleCI (by clicking on the green arrow above), which isn't happening right now. (Or you could run `make` on your computer and make sure it's created.)

Androbin:
@shardulc Just needed to fix a reference.

Androbin:
Many constants were previously overwritten with layers of defaults.
I needed a moment to figure out which one should get set respectively.

Androbin:
I ran a quick `diff` over `master/build` and `gci/build` and apart from the non-deterministic ordering of a few lines, they seem to match up.

Androbin:
I have moved all constants all JS files including imports/exports to the config file and tested everything.
@sushain97 I hope this is to your liking.

Androbin:
@shardulc Would you prefer `config-custom.conf` or `custom.conf` after all?
In addition to `string`, `int`, `bool` and `list` (of `string`), I will add support for `list` of `int`.

README.md
```diff
@@ -24,7 +24,7 @@ Setup
 -------
 
 1. Copy `config.conf.example` to `config.conf` and edit it.
-1. Then type `make`.
+2. Then type `make`.
```
sushain97: Is there a reason you changed this?

assets/js/analyzer.js
```diff
@@ -24,7 +23,7 @@ if(modeEnabled('analyzation')) {
         });
 
         $('#morphAnalyzerInput').keydown(function (e) {
-            if(e.keyCode === ENTER_KEY_CODE && !e.shiftKey) {
+            if(e.keyCode === config.ENTER_KEY_CODE && !e.shiftKey) {
```
sushain97: This shouldn't be configurable...

assets/js/persistence.js
```diff
@@ -3,14 +3,12 @@
     populateSecondaryGeneratorList, isSubset, handleNewCurrentLang */
 /* global srcLangs:true, dstLangs:true, recentSrcLangs: true, recentDstLangs:true, curSrcLang:true, curDstLang:true, locale:true */
 
-var URL_PARAM_Q_LIMIT = 1300,
```
sushain97: This shouldn't be configurable either. The things that are technical limitations and don't make sense to configure, shouldn't be configurable.

assets/js/translator.js
```diff
@@ -7,24 +7,10 @@ var recentSrcLangs = [], recentDstLangs = [];
...
-    TRANSLATION_LIST_WIDTH = 650,
-    TRANSLATION_LIST_ROWS = 8,
-    TRANSLATION_LIST_COLUMNS = 4,
-    TRANSLATION_LISTS_BUFFER = 50;
-
-var INSTANT_TRANSLATION_URL_DELAY = 500,
-    INSTANT_TRANSLATION_PUNCTUATION_DELAY = 1000,
-    INSTANT_TRANSLATION_DELAY = 3000;
-
-var PUNCTUATION_KEY_CODES = [46, 33, 58, 63, 47, 45, 190, 171, 49]; // eslint-disable-line no-magic-numbers
```
sushain97: Should not be configurable.

assets/js/util.js
```diff
@@ -1,22 +1,7 @@
 /* @flow */
 /* exported sendEvent, modeEnabled, filterLangList, getURLParam, onlyUnique, isSubset, safeRetrieve, callApy, apyRequestTimeout, isURL */
-/* exported SPACE_KEY_CODE, ENTER_KEY_CODE, HTTP_OK_CODE, HTTP_BAD_REQUEST_CODE, XHR_LOADING, XHR_DONE */
 /* global config, persistChoices, iso639Codes, iso639CodesInverse, populateTranslationList, showTranslateWebpageInterface */
 
-var SPACE_KEY_CODE = 32, ENTER_KEY_CODE = 13,
```
sushain97: None of these should be configurable. They can't be different in any sane world.

custom.conf
```diff
@@ -0,0 +1,35 @@
+[PERSISTENCE]
+URL_PARAM_Q_LIMIT = 1300
+DEFAULT_EXPIRY_HOURS = 24
+
+[TRANSLATOR]
```
sushain97: `'TRANSLATION'` for consistency

tools/conf-dtypes.txt
```diff
@@ -0,0 +1,63 @@
+[APY]
```
sushain97: While this file is OK, I'd much rather we not have it at all. It seems to be largely possible to automatically differentiate between these types.

tools/read-conf.py
```diff
@@ -7,57 +7,82 @@
...
+            parts = line.split('|')
+            parts = [part.strip() for part in parts]
 
-        'SHOW_NAVBAR'                    : conf_APY.getboolean('SHOW_NAVBAR', fallback=False),
+            if len(parts) == 1:
+                section = parts[0][1:-1]
+                dtypes[section] = {}
+            elif len(parts) == 2:
+                key = parts[0]
+                dtype = parts[1]
```
sushain97: `key, dtype = parts`?

README.md
Androbin: See commit dfe098bc8d54dcf252a4fc742e3bfc3f659faaa1 but I removed the other line in commit 7c9f41a43e14b5ca6e9f619b2a6c190ff5c7ac19

assets/js/persistence.js
Androbin: But this limitation is not technical, it's practical.

assets/js/analyzer.js
Androbin: What constants shall I move back to JS?

custom.conf
Androbin: I named it after `translator.js`

README.md
sushain97: OK, let's revert it then?

tools/conf-dtypes.txt
Androbin: But there is a chance that a number is meant to be a string. Also, a flawky config file should not mis-determine the dtype of the value.

assets/js/persistence.js
sushain97: No, it's technical. Some browsers have limitations on maximum URL length.

tools/read-conf.py
Androbin: Sure

assets/js/analyzer.js
sushain97: I think you can figure out the correct set.

tools/conf-dtypes.txt
sushain97: Where should a number be interpreted as a string? That just sounds like a misleading configuration variable that should be replaced with an enum or something.

README.md
Androbin: Why, shouldn't it be 1. and 2. instead of 1. and 1. ?

assets/js/persistence.js
Androbin: Yeah, but there is no definite maximum. I'll revert this anyway.

sushain97:
The latter never requires changing irrelevant lines for inserting an item,
i.e. smaller diffs.

On Dec 3, 2017 3:09 PM, "Robin Richtsfeld" <notifications@github.com> wrote:

*@Androbin* commented on this pull request.
------------------------------

In README.md
<https://github.com/goavki/apertium-html-tools/pull/216#discussion_r154535110>
:

> @@ -24,7 +24,7 @@ Setup
 -------

 1. Copy `config.conf.example` to `config.conf` and edit it.
-1. Then type `make`.
+2. Then type `make`.

Why, shouldn't it be 1. and 2. instead of 1. and 1. ?

—

<https://github.com/goavki/apertium-html-tools/pull/216#discussion_r154535110>,
or mute the thread

tools/conf-dtypes.txt
Androbin: The line  `CONSTANT_NAME = 42` would be "automatically differentiated" to `int` even if `CONSTANT_NAME` is a `string` and should be `"42"`

sushain97: I don't deny that it's theoretically possible. Rephrased: Where should a number be interpreted as a string [in a reasonable setting with our set of options or a reasonable extension]?

sushain97: The balance to be stricken here is between being strict and type safe or being more maintainable and having less repetition. Unless there's a good argument for the former, the latter is preferable.

Androbin: @shardulc initially wanted a dict in `read-conf.py` but I suggested to have a separate file for that.

sushain97: I prefer a dict in `read-conf.py` over this as well and over both I prefer not having a list at all. The example configuration files should serve as documentation for the options. If one is unclear, a comment would be allay the concern.

sushain97: The separate file also introduces a new syntax which is not ideal.

sushain97: Consider for example how Django does their `settings.py`. There's no exhaustive list of options.

README.md
sushain97: The latter never requires changing irrelevant lines for inserting an item,
i.e. smaller diffs.

Androbin:
> It seems to be largely possible to automatically differentiate between these types.

I'll try to match the types using the following regular expressions (with leading and trailing space already trimmed):

Pretty straightforward:
`bool` -> `True|False`
`int` -> `[0-9]+`
`int[]` -> `[0-9]+([, ]+[0-9]+)+`

But what about this?
`string` -> `.+`
`string[]` -> `.+([, ]+.+)+`

sushain97:
Does it make sense for string to be whatever doesn't match any of the above?

shardulc:
Here's where the conflicts come in: booleans have to accomodate 'True', 'Yes', 'On', and '1' as per `configparser`, but all of these can be strings and one could be an integer... plus we wouldn't want just integers, we'd want to detect all numbers, so it's a little more complex than the regex @Androbin suggested.

Androbin:
@shardulc And what about strings with commas inside? I think this goes a bit out of hand here.

shardulc:
I agree with @Androbin. @sushain97, I think it is best to explicitly say what type each config option has. But I would definitely prefer this to be a dictionary inside `read-conf.py` because like @sushain97 said, the extra file is just another file format which we need to document.

sushain97:
Hmm, OK. This does seem too tricky to be worth it. In that case, a dictionary at the top of `read-conf` probably makes the most sense where the keys are config item names and the values are parsing functions.

Androbin:
@shardulc @sushain97 Instead of a dict, what about a list of names for each type?

shardulc:
That's just confusing. It makes sense to have them in the same order in the config files and in `read-conf.py`.

sushain97:
+ diffs aren't as clear

Androbin:
Okay, gimme a moment.

Androbin:
BTW. Would you prefer a dict to `"string"` or `get_string`?

sushain97:
> BTW. Would you prefer a dict to "string" or get_string?

The function itself, i.e. the latter. This prevents having to case on the string later.

config.conf.example
```diff
@@ -2,12 +2,13 @@
 HTML_URL = https://www.apertium.org
 APY_URL = https://www.apertium.org/apy
 
-#SUBTITLE =
-#SUBTITLE_COLOR =
+SUBTITLE
```
sushain97: Why was this changed? The previous construction was clearer.

sushain97:
@shardulc what do you think makes for a good file name? `constants.conf`?

assets/js/util.js
```diff
@@ -7,16 +7,6 @@ var SPACE_KEY_CODE = 32, ENTER_KEY_CODE = 13,
     HTTP_OK_CODE = 200, HTTP_BAD_REQUEST_CODE = 400,
     XHR_LOADING = 3, XHR_DONE = 4;
 
-var TEXTAREA_AUTO_RESIZE_MINIMUM_WIDTH = 768,
-    BACK_TO_TOP_BUTTON_ACTIVATION_HEIGHT = 300,
-    APY_REQUEST_URL_THRESHOLD_LENGTH = 2000, // maintain 48 characters buffer for generated parameters
```
shardulc: I don't know what the comment is about but you should probably preserve it in the config file.

tools/read-conf.py
```diff
@@ -7,57 +7,105 @@
...
+
 
 def check_config(conf, result):
     # Some error checking:
     for section in conf.sections():
-        if section not in ['APY', 'REPLACEMENTS']:
+        if section not in ['APY', 'PERSISTENCE', 'REPLACEMENTS', 'TRANSLATOR', 'UTIL']:
             raise configparser.Error("\nUnknown section [%s]" % (section,))
 
+    # TODO: either remove or check for all sections
```
shardulc: Instead of checking the result after it's been generated, you could: 1. do `check_config` before you populate `results` and 2. check whether each item in all sections of `config` is present in `dtypes`.

shardulc: More helpful error messages throughout this code would be good, too.

tools/read-conf.py
```diff
@@ -7,57 +7,105 @@
...
 
     return True
 
-def load_conf(filename):
-    conf = configparser.ConfigParser()
-    with open(filename, 'r') as f:
-        conf.read_file(f)
-    conf_APY = conf['APY']
-    result = {
-        'HTML_URL'                       : conf_APY.get('HTML_URL', fallback="http://www.apertium.org"),
```
shardulc: You forgot about the fallbacks this time :)

tools/read-conf.py
```diff
@@ -84,6 +140,7 @@ def print_keyval(result, args):
 if __name__ == "__main__":
     parser = argparse.ArgumentParser(description='Load config, print stuff')
     parser.add_argument('-c', '--config', default='config.conf', help='Config file name (default: config.conf)')
+    parser.add_argument('-C', '--custom', default='custom.conf', help='Customization file name (default: custom.conf)')
```
shardulc: @sushain97 What about `internal-constants.conf`? (@Androbin don't forget to rename variables once we decide.)

config.conf.example
Androbin: At first, I moved the fallback into the custom format file. But when I turned it into a conf file, I set the initial values directly to the fallbacks. I will move the non-`None` fallbacks back into `read-conf.py`, a simple `dict` should do.

assets/js/util.js
Androbin: Sure.

tools/read-conf.py
Androbin: I did not write either the test nor the error messages. But I will try to implement your suggestions.

tools/read-conf.py
Androbin: I had them in the custom format file but when I turned it into a config file, I set the initial values to the fallbacks. I will add a `dict` for all non-`None` fallbacks in `read-conf.py`

tools/read-conf.py
Androbin: But "internal" sounds like the path should not be configurable at all.

assets/js/util.js
Androbin: I think `2000 = 2^11 - 48` btw.

Androbin:
@shardulc @sushain97 I hope this approach is of both your liking. I do now list and describe all constants in `read-conf.py` and can now iterate over this `dict` instead of iterating over the values in the `conf` files.

assets/js/util.js
sushain97: I don't actually think it should be configurable.

tools/read-conf.py
sushain97: What's wrong with `constants.conf`? Seems OK to me.

tools/read-conf.py
```diff
@@ -7,58 +7,115 @@
...
+    array = get_string_array(conf, key)
+    return fallback if array is None else [int(x) for x in array]
+
+
+# Constant descriptors
+constants = {
+    # section name
+    'APY': {
+        # CONSTANT_NAME: [parser]
+        # CONSTANT_NAME: [parser, fallback]
```
sushain97: This should not be an list, a tuple is a far far better choice.

assets/js/util.js
Androbin: Okay

tools/read-conf.py
Androbin: Well, `config.conf` contains "constants" as well.

tools/read-conf.py
Androbin: Sure

sushain97:
@shardulc this looks mostly OK to me aside from the tuple/list comment above. I think if you can verify that it doesn't break any of the modes, its safe to approve the task. We can merge after we figure out naming of the second config file and take a last close look at the code.

tools/read-conf.py
sushain97: Fair enough. Hmm... `advanced-config.conf`?

Androbin:
@sushain97 I can't quite figure out how to rewrite it using tuples.
Python's unpacking rules are somewhat mean to me.

sushain97:
What's the error? You need to use (firstitem, ) to make a tuple with one
item.

Androbin:
> You need to use `(firstitem, )` to make a tuple with one
Oh, didn't know that. Thanks, will try again.

tools/read-conf.py
```diff
@@ -67,6 +124,12 @@ def load_conf(filename):
         'ANALYZERS': None,
         'TAGGERS': None
     }
+
+    for section_name, section_desc in constants.items():
+        for constant_name, constant_desc in section_desc.items():
+            dtype, fallback, *_ = (*constant_desc, None)
```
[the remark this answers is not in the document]
shardulc: Um why not just `dtype, fallback = constant_desc`?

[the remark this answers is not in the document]
Androbin: Does not work when no fallback is specified.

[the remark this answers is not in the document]
Androbin: But I got an idea. I will add a function to pad the tupel with `None`.

sushain97:
This just screams hack. I can't think of an elegant way to destructure both
a pair and singleton simultaneously in Python so I suggest just going with
the more verbose conditional option.

Androbin:
@sushain97 What you think about my padding approach?

shardulc:
@sushain97 @Androbin I think I have an elegant solution. What if:
 - every single constant in `read-conf.py` had a fallback whose value is the default for html-tools (i.e. the values that we know to work)
 - every single line in the file `advanced-config.conf` is commented, and some text explains how you would customize the constants by uncommenting and changing them

Androbin:
@shardulc Do you propose to dissolve `config.conf.example` into the fallbacks in `read-conf.py` ?

sushain97:
I actually don't like that approach too much because it requires editing a
Python file to change the interface. I think it makes everything a little
too connected and feels very magical.

I think the verbose method of figuring out whether there's a fallback or
not is fine. It's like 2 lines of code extra.

shardulc:
> requires editing a Python file to change the interface
No, I edited my comment to make it more clear.

Androbin:
@shardulc What is keeping us from combining `config,conf` and `custom.conf` anymore then?

sushain97:
> No, I edited my comment to make it more clear.

A developer would have to modify the Python to change how the interface looks by default. This seems not great to me with regards to separation of concerns.

Androbin:
@sushain97 As far as I understand @shardulc we'd essentially dissolve `config.conf.example` into the fallbacks in `read-conf.py`

sushain97:
Well, I think he means dissolving `advanced-config.conf` into `read-conf.py`. We can't remove `config.conf.example`, it would seriously break backwards compatibility.

shardulc:
> dissolving `advanced-config.conf` into `read-conf.py`

Yes, that's what I meant.

> A developer would have to modify the Python

I was thinking only goavki would modify the Python when required to tweak constants. If downstream users want to tweak constants, they'd be recommended to create/change `advanced-config.conf` instead so that they can keep their tweaks even when we push different fallbacks. (Changing the conf file does not require a developer either.)

sushain97:
> I was thinking only goavki would modify the Python when required to tweak constants.

Right. This is my issue because we lose separation of concerns. Right now the Python is just a tool without any semantic meaning (it has fallbacks but those are duplicated in `config.conf.example`). If we implemented this idea, this would no longer be the case. We could store the defaults in the JS and hydrate the configuration with those defaults when they're missing, i.e. `advanced-config.conf.example` would have commented out constants and it could be moved to `advanced-config.conf`.

Either way, I think we should approve @Androbin's task and figure this out amongst ourselves. We should have asked for a concrete plan as one task and implementation as another since there are lots of ways to do this :)

shardulc:
Ah, OK. @Androbin I'm approving your task then.

Androbin:
Any updates on this yet?

shardulc:
@Androbin No 'updates', but discussion is [over here](https://github.com/goavki/apertium-html-tools/issues/158#issuecomment-349368088). If you want to resolve those comments, you can do it as part of another task. (It won't be enough for one task because in my opinion it's only a few changes, you can pair it with another easy issue.)

sushain97:
I haven't had a chance to look at it carefully yet. As part of the flow
updates, the approach to this will have to change a bit so let's not make
any changes yet.

Androbin:
@sushain97 When the `flow` changes are done, you can give an outline of these desired changes.

Androbin:
> As part of the flow updates, the approach to this will have to change a bit so let's not make any changes yet.

@sushain97 Could you give an outline of these desired changes.

PR #215, opened 2017-12-01
File download in IE10+ fixing #97

myfrom:
I added a non-breaking workaround for issue #97 in IE10 and 11.  
  
I tested both of them in virtual machine at Sauce Labs, logs for [IE10](https://saucelabs.com/beta/tests/5b3dda1b469941da93b32821ce818a62/metadata) and [IE11](https://saucelabs.com/beta/tests/f3bd41c0197b44cc9a02aabd83d5e30e/metadata).  
Also checked Chrome 62 locally and can confirm it works as before the change.

This is a part of [Google Code-In task](https://codein.withgoogle.com/dashboard/task-instances/5358971881783296/)

(Screenshot from IE)
0003screenshot

Issue #97, opened 2017-01-17
Automatically download translated file for Safari, IE, etc.


unhammer:
Safari users report documents getting back all garbled

![garble](https://cloud.githubusercontent.com/assets/56868/22023743/c1483408-dcc7-11e6-94f1-7cb4ab095fe1.png)

[tull_test.txt](https://github.com/goavki/apertium-html-tools/files/711250/tull_test.txt)

(moved from https://github.com/goavki/apertium-apy/issues/48 )

unhammer:
We might want to temporarily give Safari users a warning about upgrading to a less [buggy](https://nolanlawson.com/2015/06/30/safari-is-the-new-ie/) browser.

sushain97:
I can confirm this on Safari 9.

I suspect it has something to do with https://github.com/goavki/apertium-html-tools/blob/master/assets/js/translator.js#L644 but I can't find any reliable analysis/solutions from cursory Google searching.

unhammer:
I tried epiphany-browser (which sometimes shares Safari's quirks?), and it also fails. 

Problem 1: we check file.type, which is non-standard: https://developer.mozilla.org/en-US/docs/Web/API/IDBMutableFile/type 
We already check for mime type on the web server (but we show server-mime-failures as "translation not available"!)

Problem 2: I tried using the html "accept" attrib instead: http://sprunge.us/JWDR but epiphany at least sends a file that APY detects as "application/octet-stream" – I don't yet know if that's because the Content-Type of the request is such, or if the file itself differs.

sushain97:
@Problem 1: is it not https://developer.mozilla.org/en-US/docs/Web/API/Blob/type ?

unhammer:
Oh, right. When I log the `droppedFile` in epiphany, it tells me it actually has a `.type`, it's just set to the empty string … I suppose we should allow empty strings, since that at least means the browser hasn't explicitly given a different mime type.

unhammer:
29b009748a0c085c5e770aea6bec70cea4c6224f adds "" as an accepted MIME type js-side (server will, and should, do the real check anyway) and adds accept attribute. Epiphany now lets me drop a document and translate it, but when I click the download link, I don't get an .odt, but it sends me to an address `blob:null/a0594597-30c7-480e-a968-846d140085e0` containing what I suppose is the odt contents:
[blob.txt](https://github.com/goavki/apertium-html-tools/files/720525/blob.txt)

unhammer:
[...]

PR #219, opened 2017-12-03
Add file download workaround for Safari (progress #97)

myfrom:
I added another workaround to file downloading, this time for Safari. I tested it on Chrome 62 and Safari 9 (on cloud machine) but as it's not as clean workaround as #215 I'd recommend that someone with access to more local browsers would test this. The file now downloads and is correctly recognised but downloads without extension type and with name 'Unknown' but I can't find any better way to do that.
I also updated the English string to be more specific in current situation but I didn't want to break anything so I left other languages untouched.

[Logs for Safari 9](https://saucelabs.com/beta/tests/7c83ff01d5b348cb82612ae835ae0834)
0003screenshot

This is a part of [GCI task](https://codein.withgoogle.com/dashboard/task-instances/5358971881783296/)

### DIFF
--- a/assets/js/translator.js
+++ b/assets/js/translator.js
@@ -740,7 +740,6 @@ function translateDoc() {
                         });
                     }
                     else if(this.readyState === XHR_DONE && xhr.status === HTTP_OK_CODE) {
-                        downloadBrowserWarn();
                         $('div#fileUploadProgress').parent().fadeOut('fast');
                         $('div#fileLoading').fadeOut('fast', function () {
                             if(typeof window.navigator.msSaveBlob !== 'undefined') {
@@ -760,7 +759,6 @@ function translateDoc() {
                                         .click(function () {
                                             window.location.href = reader.result.replace(/^data:[^;]*;/, 'data:attachment/file;');
                                         });
-                                    $('#fileDownloadBrowserWarning').show();
                                 };
                                 reader.readAsDataURL(xhr.response);
                             }
@@ -956,16 +954,6 @@ function showTranslateWebpageInterface(url, ignoreIfEmpty) {
         window.location.hash = 'webpageTranslation';
         translateWebpage(ignoreIfEmpty);
     });
-}
-
-function downloadBrowserWarn() {
-    if(typeof $bu_getBrowser == 'function') { // eslint-disable-line camelcase
-        var detected = $bu_getBrowser();
-        // Show the warning for (what bu calls) "niche" browsers and Safari, but not Chromium:
-        if(detected.n.match(/^[xs]/) && !(navigator.userAgent.match(/Chromium/))) {
-            $('#fileDownloadBrowserWarning').show();
-        }
-    }
 }
 
 function detectLanguage() {
--- a/assets/js/translator.js
+++ b/assets/js/translator.js
@@ -743,11 +743,20 @@ function translateDoc() {
                         downloadBrowserWarn();
                         $('div#fileUploadProgress').parent().fadeOut('fast');
                         $('div#fileLoading').fadeOut('fast', function () {
-                            var URL = window.URL || window.webkitURL;
-                            $('a#fileDownload')
-                                .attr('href', URL.createObjectURL(xhr.response))
-                                .attr('download', fileName)
-                                .fadeIn('fast');
+                            if(typeof window.navigator.msSaveBlob !== 'undefined') {
+                                $('a#fileDownload')
+                                    .click(function () {
+                                        window.navigator.msSaveBlob(xhr.response, fileName);
+                                    })
+                                    .fadeIn('fast');
+                            }
+                            else {
+                                var URL = window.URL || window.webkitURL;
+                                $('a#fileDownload')
+                                    .attr('href', URL.createObjectURL(xhr.response))
+                                    .attr('download', fileName)
+                                    .fadeIn('fast');
+                            }
                             $('span#fileDownloadText').text(getDynamicLocalization('Download_File').replace('{{fileName}}', fileName));
                             $('button#translate').prop('disabled', false);
                             $('input#fileInput').prop('disabled', false);
--- a/assets/js/translator.js
+++ b/assets/js/translator.js
@@ -25,7 +25,6 @@ var PUNCTUATION_KEY_CODES = [46, 33, 58, 63, 47, 45, 190, 171, 49]; // eslint-di
     getDynamicLocalization, locale, ajaxSend, ajaxComplete, localizeInterface, filterLangList, cache, readCache, iso639Codes,
     callApy, apyRequestTimeout, isURL */
 /* global SPACE_KEY_CODE, ENTER_KEY_CODE, HTTP_OK_CODE, XHR_LOADING, XHR_DONE, HTTP_OK_CODE, HTTP_BAD_REQUEST_CODE */
-/* global $bu_getBrowser */
 
 if(modeEnabled('translation')) {
     $(document).ready(function () {
--- a/assets/js/translator.js
+++ b/assets/js/translator.js
@@ -749,9 +749,8 @@ function translateDoc() {
                                         window.navigator.msSaveBlob(xhr.response, fileName);
                                     });
                             }
-                            else if(/.*(?!chrome|android).*(version\/[^10,11]|Mobile).*safari/i.test(window.navigator.userAgent) &&
+                            else if(/.*(?!chrome|android).*(version\/\d\.|Mobile).*safari/i.test(window.navigator.userAgent) &&
                                 !/chrome/i.test(window.navigator.userAgent)) {
-                                // When a new version of Safari comes out ^^ need to update
                                 // Safari <9 + Mobile file download workaround
                                 var reader = new FileReader();
                                 reader.onload = function () {
--- a/assets/js/translator.js
+++ b/assets/js/translator.js
@@ -744,12 +744,14 @@ function translateDoc() {
                         $('div#fileUploadProgress').parent().fadeOut('fast');
                         $('div#fileLoading').fadeOut('fast', function () {
                             if(typeof window.navigator.msSaveBlob !== 'undefined') {
+                                // IE file download workaround
                                 $('a#fileDownload')
                                     .click(function () {
                                         window.navigator.msSaveBlob(xhr.response, fileName);
                                     });
                             }
                             else if(/^((?!chrome|android).)*safari/i.test(window.navigator.userAgent)) {
+                                // Safari file download workaround
                                 var reader = new FileReader();
                                 reader.onload = function () {
                                     $('a#fileDownload')
@@ -761,6 +763,7 @@ function translateDoc() {
                                 reader.readAsDataURL(xhr.response);
                             }
                             else {
+                                // File download for modern browsers
                                 var URL = window.URL || window.webkitURL;
                                 $('a#fileDownload')
                                     .attr('href', URL.createObjectURL(xhr.response))
--- a/assets/js/translator.js
+++ b/assets/js/translator.js
@@ -762,14 +762,12 @@ function translateDoc() {
                                 };
                                 reader.readAsDataURL(xhr.response);
                             }
-                            else {
-                                // File download for modern browsers
-                                var URL = window.URL || window.webkitURL;
-                                $('a#fileDownload')
-                                    .attr('href', URL.createObjectURL(xhr.response))
-                                    .attr('download', fileName);
-                            }
-                            $('a#fileDownload').fadeIn('fast');
+                            // File download for modern browsers and right-click to save
+                            var URL = window.URL || window.webkitURL;
+                            $('a#fileDownload')
+                                .attr('href', URL.createObjectURL(xhr.response))
+                                .attr('download', fileName)
+                                .fadeIn('fast');
                             $('span#fileDownloadText').text(getDynamicLocalization('Download_File').replace('{{fileName}}', fileName));
                             $('button#translate').prop('disabled', false);
                             $('input#fileInput').prop('disabled', false);
--- a/assets/js/translator.js
+++ b/assets/js/translator.js
@@ -747,8 +747,7 @@ function translateDoc() {
                                 $('a#fileDownload')
                                     .click(function () {
                                         window.navigator.msSaveBlob(xhr.response, fileName);
-                                    })
-                                    .fadeIn('fast');
+                                    });
                             }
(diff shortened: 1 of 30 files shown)


sushain97:
Yeah, I can't really review this current iteration because of the massive diff :) @mention me when it's sane again.

assets/js/translator.js
```diff
@@ -743,11 +743,32 @@ function translateDoc() {
...
+                            if(typeof window.navigator.msSaveBlob !== 'undefined') {
+                                $('a#fileDownload')
+                                    .click(function () {
+                                        window.navigator.msSaveBlob(xhr.response, fileName);
+                                    })
+                                    .fadeIn('fast');
+                            }
+                            else if(/^((?!chrome|android).)*safari/i.test(window.navigator.userAgent)) {
+                                var reader = new FileReader();
+                                reader.onload = function (e) {
```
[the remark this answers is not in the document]
sushain97: You can make it `_e`

assets/js/translator.js
```diff
@@ -743,11 +743,32 @@ function translateDoc() {
...
+                                        $('#fileDownloadBrowserWarning').show();
+                                }
+                                reader.readAsDataURL(xhr.response);
+                            }
+                            else {
+                                var URL = window.URL || window.webkitURL;
+                                $('a#fileDownload')
+                                    .attr('href', URL.createObjectURL(xhr.response))
+                                    .attr('download', fileName)
+                                    .fadeIn('fast');
```
sushain97: Looks like we're always `.fadeIn('fast')`ing? Can we move that out of the conditional block?

assets/strings/eng.json
```diff
@@ -30,7 +30,7 @@
     "Format_Not_Supported": "Format not supported",
     "Translate_Webpage": "Translate a webpage",
     "Supported_Formats": "Supported formats: .txt, .rtf, .odp, .ods, .odt (LibreOffice/OpenOffice), .xlsx, .pptx, .docx (Microsoft Office 2003 and newer). <br/><i>Word 97 .doc files and pdf's are not supported.</i>",
-    "Download_Browser_Warning": "Your browser might not handle downloads correctly. Either right-click and save the link with the correct filename or open this page in <a href=\"https://www.mozilla.org/firefox/\">Firefox</a>.",
+    "Download_Browser_Warning": "Your browser does not handle downloads correctly so downloaded file will save as \"Unknown\", you might want to rename it. Alternatively you can open this page in a browser that handles downloads properly, such as <a href=\"https://www.mozilla.org/firefox/\">Firefox</a>.",
```
sushain97: Hmmm... I actually think leaving it at the old state might be OK because having vague messaging is not that bad and prevents having to re-translate every time something specific changes.

myfrom: Not sure if after that change you still can right click to save in Safari. Would need someone with this browser installed to try.

sushain97: Ah, that would be unfortunate and necessitate a change. I can check. If the change is required. all the other translation strings need to be deleted.

assets/js/translator.js
```diff
@@ -743,11 +743,32 @@ function translateDoc() {
                         downloadBrowserWarn();
                         $('div#fileUploadProgress').parent().fadeOut('fast');
                         $('div#fileLoading').fadeOut('fast', function () {
-                            var URL = window.URL || window.webkitURL;
-                            $('a#fileDownload')
-                                .attr('href', URL.createObjectURL(xhr.response))
-                                .attr('download', fileName)
-                                .fadeIn('fast');
+                            if(typeof window.navigator.msSaveBlob !== 'undefined') {
+                                $('a#fileDownload')
```
sushain97: Could you add a one-liner comment to each of these conditional blocks stating what browser the support is for?

assets/strings/eng.json
sushain97: Hm, you're right. It seems possible that a hybrid approach would work (i.e. adding `href` and `download` anyway).

sushain97: Also, would this problem with Safari not also affect IE?

sushain97:
Appears to be a minor conflict to resolve as well :)

assets/strings/eng.json
myfrom: Yeah, it's impossible to download it in IE using right-click but my IE workaround uses a different method (`msSaveBlob` function implemented only by Microsoft) than Safari and so it saves with correct file name. And the warning never shows in IE.

Back to Safari, I'll do what you said and revert the string changes.

sushain97: > Yeah, it's impossible to download it in IE using right-click but my IE workaround uses a different method (msSaveBlob function implemented only by Microsoft) than Safari and so it saves with correct file name.

I see. Do you think it would hurt to make it possible to download via right-click as well? Users might be accustomed to doing different things and if it doesn't cost us much in terms of LOC/maintainability, I'm fine doing it.

> And the warning never shows in IE.

Great.

> Back to Safari, I'll do what you said and revert the string changes.
 
Perfect.

myfrom:
I will check that with IE, but in ~2hours

myfrom:
Should be OK now. However, I don't have time to fight with cloud machine to test it today.

sushain97:
OK, let's distribute our testing.

We need to test 2 things
- clicking on link downloads
- right click + appropriate menu item downloads

On multiple browsers
- Chrome
- IE (10+)
- Edge
- Firefox
- Safari

@shardulc which of the above can you test? I can test the rest, probably.

shardulc:
@sushain97 I can test Firefox and Chrome on my laptop and on my phone.

sushain97:
on IE:

click works, save target as does nothing?

on Edge:

click works, save target as has a weird filename (maybe not fixable?):

image

sushain97:
on Safari:

download file as right click works, clicking downloads *twice*

shardulc:
On desktop Firefox: works perfectly.
On desktop Chrome: works perfectly, except that the saved file's default extension is .html.

shardulc:
On mobile Firefox: it says "Translation not available", no link.
On mobile Chrome: works fine, but there is no option to choose the name of
the saved file (I think this is true for all mobile downloads though).

myfrom:
- IE: Probably bug, will look into it
- Edge: Probably unfixable, because of how browser handles downloads
- Desktop Chrome: Can't reproduce on Chrome 62, Chrome OS. What file did you use to test?
- Mobile Firefox: I can't test it but seems like a problem with the part of the service that I haven't modified.
- Mobile Chrome: Weird, should download with a normal file name; but the fact that user can't change it is expected.
- Safari: A tricky one: Safari 10+ on desktop supports `download` attribute in `<a>` so downloads it once, and then my workaround kicks in and downloads it again. Safari on mobile and <9 [don't support](https://caniuse.com/#feat=download) that attribute and should download only once. It would be impossible to change this without a working way to detect Safari version which I can't find.

I'll get into bugs when I'm back from school

sushain97:
> It would be impossible to change this without a working way to detect Safari version which I can't find.

image


image

myfrom:
Oh, thanks.

sushain97:
> Edge: Probably unfixable, because of how browser handles downloads

Fair. All I ask is that an issue is opened for it after this one is closed. We can label it blocked or something.

shardulc:
> Can't reproduce on Chrome 62, Chrome OS. What file did you use to test?

Hmm yes, I used a text file without any extension to test. When I test with the same file `.txt` it downloads as `.txt` too.

assets/js/translator.js
```diff
@@ -744,19 +744,31 @@ function translateDoc() {
...
-                                    .fadeIn('fast');
+                                    });
                             }
-                            else {
-                                var URL = window.URL || window.webkitURL;
-                                $('a#fileDownload')
-                                    .attr('href', URL.createObjectURL(xhr.response))
-                                    .attr('download', fileName)
-                                    .fadeIn('fast');
+                            else if(/.*(?!chrome|android).*(version\/[^10,11]|Mobile).*safari/i.test(window.navigator.userAgent) && !/chrome/i.test(window.navigator.userAgent)) {
```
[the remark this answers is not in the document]
sushain97: Wow this sucks.

sushain97:
OK, I should've just said forget old Safari but now that you've done it 👏

myfrom:
Thanks 😄 
But it requires to update the code when new version comes out or this bug will come back. (Add new version number to RegExp, commented that on line 755).

I checked IE10 and I don't have a problem with right-click, only need to change file name and add extension but I can't avoid that.

I will open that issue for Edge as you said.

I checked in mobile/desktop Chrome locally + IE10 in cloud. I also have done some testing on RegExp that is meant to detect Safari by trying it on user agents that I found in the web.

sushain97:
> But it requires to update the code when new version comes out or this bug will come back. (Add new version number to RegExp, commented that on line 755).

:( It seems like instead of doing not 10, 11 we could do 1-digit number only? (somewhat hacky, yes)

@shardulc let's do another round of testing.

|    | Left Click | Right Click |
| - | ----- | ----------- |
| Safari |  | |
| Chrome |  | |
| IE |  | |
| Edge |  | |
| Firefox |  | |
| Chrome Mobile |  | |
| Safari Mobile |  | |

Please update the above table with ✓ if it works and comment otherwise. I don't really care much about any browsers not on the above list or old versions. We're not Google, supporting everything isn't the goal.

myfrom:
Chrome left click works for sure.

That detection for one digit should go through...
I'll try to update it from mobile, give me a second

myfrom:
Done, but now it would be good if someone could check older Safari (9 or earlier) too to see if the regex is still valid

sushain97:
I checked three browsers and added one comment with a problem I found above.

One other minor issue:
- I translate a file on a message with the browser warning.
- Drag another file onto the page
- Download link for old file fades out but browser warning stays! (makes it jump up)
- New download link appears (makes browser warning jump)

Either we can make download link fade out in visibility only (opacity 0; still takes up space) or we can make both browser warning and download link disappear and appear. It would be great if you could easily fix this since you've been knee deep in this code!

Regardless, we're really close. I need to check on Windows browsers later today and @shardulc can verify mobile sanity.

myfrom:
How about removing this warning all together? The workarounds should make it work everywhere now.

Whatever you choose, let me know and I'll get this done tomorrow.
Could I just ask you for another day? Sorry for the delay but we have a few exams in school this week.

sushain97:
> How about removing this warning all together? The workarounds should make it work everywhere now.

Sounds even better.

> Could I just ask you for another day? Sorry for the delay but we have a few exams in school this week.

No problem. Good luck!

BTW, updated table.

shardulc:
> I don't really care much about any browsers not on the above list

Firefox mobile, too? I use Firefox mobile. Recommenting table:

  | Left Click | Right Click | Comments
-- | -- | -- | --
Safari | ✓ | ✓ | Remove "browser might not handle..."
Chrome | ✓ | ✓ |  
IE | ✓ | ✗ | right click -> "Save Target As" does nothing
Edge | ✓ | ✓ | "Save Target As" has weird name (issue pending)
Firefox | ✓ | ✓ |  
Chrome Mobile | ✓ | ✓ | weird name, but I don't think mobile browsers let you choose anyway
Safari Mobile |   |   | the translated file (.docx, .rtf tried so far) opens as plaintext in the browser
Firefox Mobile | ✗ | ✗ | "Translation not yet available!" but this browser is recommended by the warning

sushain97:
> Firefox mobile, too? I use Firefox mobile.

But who else does? ;) 

> "Translation not yet available!"

Sounds like this requires some really ugly remote debugger.

> but this browser is recommended by the warning

Hm, I'm guessing that the warning means desktop Firefox :)

myfrom:
> IE: right click -> "Save Target As" does nothing

Can't reproduce on IE10, IE11 on cloud machine using sample .txt file

> Firefox Mobile: "Translation not yet available!" but this browser is recommended by the warning

This looks like an issue of code that I haven't touched. Could you also try the version from master?

> Safari Mobile: the translated file (.docx, .rtf tried so far) opens as plaintext in the browser

Can't do further tests on my end. I synthetically tested user agents from Safari on iOS and my workaround recognises it properly and should apply the fix. I guess the reason is somewhere in Safari, as at school I asked my friend with an iPhone to try a different site which downloads files and it presents the same behavior. Also "save target as" functionality is unavailable completely on Safari Mobile when I tested on my friend's phone.

> Remove "browser might not handle..."

Done, removed every instance of this string that I found

shardulc:
> This looks like an issue of code that I haven't touched.

I know, I'm just including it here so that we have complete knowledge of what the current status is. `master` has the same problem.

> I guess the reason is somewhere in Safari

Yeah, in that case I don't think there will be a nice way to fix this. @sushain97 Except for IE, this task is ready for final review?

sushain97:
> Yeah, in that case I don't think there will be a nice way to fix this. @sushain97 Except for IE, this task is ready for final review?

Honestly, I think it's fine. If it works for @myfrom it might just be my machine.

Let's open a ticket for the Safari mobile being borked.

sushain97:
The code looks good to me. @shardulc could you take a look as well? After this gets merged, we need to run `cleanup.sh` and cut a release (v3.0.1).

assets/js/translator.js
```diff
@@ -740,23 +739,33 @@ function translateDoc() {
...
-                                    .fadeIn('fast');
+                                    });
                             }
-                            else {
-                                var URL = window.URL || window.webkitURL;
-                                $('a#fileDownload')
-                                    .attr('href', URL.createObjectURL(xhr.response))
-                                    .attr('download', fileName)
-                                    .fadeIn('fast');
+                            else if(/.*(?!chrome|android).*(version\/\d\.|Mobile).*safari/i.test(window.navigator.userAgent) &&
```
shardulc: A little explanation for this, perhaps?

shardulc: @sushain97 we can do this ourselves really, after merging.

myfrom: Tests User Agent against a regexp to detect if the user is on older or mobile version of Safari and then runs another test to eliminate Chrome, which also include Safari in User Agent.

Should I add there a comment? I thought that the one line below is enough

shardulc: Yes, just expand that comment a bit. The thing you just said would be perfect.

sushain97:
@myfrom do you mind opening a ticket for Safari Mobile? The way you wrote up the other one was excellent.

myfrom:
Thank you 😄 

Here it is #222

PR #226, opened 2017-12-16
Fix #220

### LINKED ISSUE #220: Make debug is borked
```
$ make debug
make: *** No rule to make target `build/index..html', needed by `debug'.  Stop.
```

I think this happened in ac38feaf174b86c4838aa447a409a7e651b657d2 @unhammer

diogoscf:
Solve make debug problem (#220)


sushain97:
Hey @diogoscf! Wow, this is all? :) I will take a look soon.

I think for the task it would be fair if you could tackle one other issue as well.

diogoscf:
I agree this was too easy

Issue #228, opened 2017-12-16
Better flow coverage


sushain97:
Not looking so great.

Should also add `flow-coverage-report` to build after it looks OK.

```
$ flow-coverage-report -i "assets/js/*.js" -x "assets/js/jquery.jsonp-2.4.0.min.js"
┌───────────────────────────┬────────────┬─────────┬───────┬─────────┬───────────┐
│ filename                  │ annotation │ percent │ total │ covered │ uncovered │
│ assets/js/analyzer.js     │    no flow │    41 % │   652 │     271 │ 381       │
│ assets/js/compat.js       │    no flow │    71 % │   366 │     260 │ 106       │
│ assets/js/generator.js    │    no flow │    38 % │   482 │     185 │ 297       │
│ assets/js/init.js         │    no flow │    47 % │   475 │     226 │ 249       │
│ assets/js/localization.js │    no flow │    52 % │  1218 │     636 │ 582       │
│ assets/js/persistence.js  │    no flow │    41 % │   879 │     369 │ 510       │
│ assets/js/sandbox.js      │    no flow │    45 % │   118 │      54 │ 64        │
│ assets/js/store.js        │    no flow │    53 % │   113 │      60 │ 53        │
│ assets/js/strict.js       │    no flow │   100 % │     1 │       1 │ 0         │
│ assets/js/translator.js   │       flow │    60 % │  2946 │    1774 │ 1172      │
│ assets/js/util.js         │       flow │    70 % │   522 │     367 │ 155       │
└───────────────────────────┴────────────┴─────────┴───────┴─────────┴───────────┘
┌─────────────────────────┬─────────────────────────────────────────┐
│ included glob patterns: │ assets/js/*.js                          │
│ excluded glob patterns: │ assets/js/jquery.jsonp-2.4.0.min.js     │
│              threshold: │ 80                                      │
│       concurrent files: │ 1                                       │
│           generated at: │ Sat Dec 16 2017 16:25:40 GMT-0500 (EST) │
│           flow version: │ 0.61.0                                  │
│      flow check passed: │ yes (0 errors)                          │
└─────────────────────────┴─────────────────────────────────────────┘
┌─────────────────────┬─────────┬───────┬─────────┬───────────┐
│ project             │ percent │ total │ covered │ uncovered │
│ apertium-html-tools │    54 % │  7772 │    4203 │      3569 │
└─────────────────────┴─────────┴───────┴─────────┴───────────┘
[...]

PR #229, opened 2017-12-16
[WIP] Better flow coverage

Androbin:
Fixing #228 @sushain97


assets/js/translator.js
```diff
@@ -666,20 +666,21 @@ function translateText(ignoreIfEmpty) {
                 clearTimeout(apyRequestTimeout);
             }
 
-            var endpoint, request;
+            var endpoint;
+            var request /*: { langpairs?: string, langpair?: string, q: string, markUnknown: string }*/ = {
```
sushain97: Should be spaces at the end of comments.

assets/js/translator.js
```diff
@@ -763,8 +764,9 @@ function translateDoc() {
                                 reader.onload = function () {
                                     $('a#fileDownload')
                                         .click(function () {
-                                            // $FlowFixMe
-                                            window.location.href = reader.result.replace(/^data:[^;]*;/, 'data:attachment/file;');
+                                            // Flow infers incorrect type ArrayBuffer
+                                            const result /*: string*/ = (reader.result/*: any*/);
```
sushain97: `const` is ES6 - do not use.

sushain97:
For this last set of CI errors I couldn't find a good solution aside from removing the space between the variable name and the comment so I guess that'll have to do.

sushain97:
If you replace the jquery stub definition with the real one, I think it brings coverage up a lot. Might have to upgrade the jQuery version to 3.x so you don't get really old defs but that should be fine.

Androbin:
Coverage is at 74% now. I was able to eliminate all but 2 errors. @sushain97 please take a quick look.

Androbin:
Not sure about possibly version-related errors.

Androbin:
Forget what I said. After adding `// @flow` to the other files, more errors appeared.

sushain97:
> Not sure about possibly version-related errors.

Yeah, I think you should upgrade jQuery:

> // Flow-typed type definitions for jQuery 3.x.x

```
build/js/jquery.min.js: build/js/.d
	curl -Ss 'http://ajax.googleapis.com/ajax/libs/jquery/1.11.3/jquery.min.js' -o $@
```

and then check that everything still functions as expected.

assets/js/util.js
```diff
@@ -190,14 +191,14 @@ function callApy(options /*: any */, endpoint /*: any */, useAjax /*: any */) {
         return $.ajax(requestOptions);
     }
 
-    return $.jsonp(requestOptions);
+    return $.ajax.jsonp(requestOptions);
```
sushain97: I don't think this is correct. `$.jsonp` is a library separate from jQuery.

assets/js/translator.js
```diff
@@ -699,9 +706,13 @@ function translateText(ignoreIfEmpty) {
 function translateDoc() {
     var validPair = pairs[curSrcLang] && pairs[curSrcLang].indexOf(curDstLang) !== -1;
 
-    var file = droppedFile ? droppedFile : undefined;
-    if($('input#fileInput')[0].files.length > 0 && $('input#fileInput')[0].files[0].length !== 0) {
-        file = $('input#fileInput')[0].files[0];
+    var file = droppedFile ? droppedFile : null;
+
+    // Flow can only infer type HTMLElement
+    var fileInput /*: HTMLInputElement */ = ($('input#fileInput')[0]/*: any */);
```
sushain97: Can this not read:

```
var fileInput = ($('input#fileInput')[0] /*: HTMLInputElement */);
```

?

assets/js/translator.js
```diff
@@ -371,7 +377,7 @@ function getPairs() {
         }
         else {
             console.warn('Translation pairs cache ' + (pairs === null ? 'stale' : 'miss') + ', retrieving from server');
-            $.jsonp({
+            $.ajax.jsonp({
```
sushain97: Same jsonp problem here as below.

assets/js/analyzer.js
```diff
@@ -241,6 +243,8 @@ function handleAnalyzeErrorResponse(xOptions, error) {
     $('#morphAnalyzerOutput').text(error).removeClass('blurred');
 }
 
+/*:: export {analyzerData, analyzers, getAnalyzers, populateAnalyzerList, populateSecondaryAnalyzerList} */
```
sushain97: Fine to have a empty line between these but I don't think all the files do.

assets/js/strict.js
```diff
@@ -1 +1,2 @@
+// @flow
```
sushain97: line break after (or just don't even add `@flow` to this file, lol)

Androbin: I know what you mean. I just wanted to make `flow coverage` happy.

assets/js/analyzer.js
Androbin: I'll make it so.

assets/js/translator.js
Androbin: Thought so, too. Apparently not. Because this isn't guaranteed to succeed, Flow doesn't permit.

assets/js/util.js
Androbin: Well, `$.jsonp` didn't resolve, while `$.ajax.jsonp` did.
Shall we stub this?

Androbin:
The module `persistence.js` is making trouble.
It overrides some variables imported from other modules.
And what is the intention behind [this line](https://github.com/goavki/apertium-html-tools/blob/master/assets/js/persistence.js#L4)?

assets/js/util.js
sushain97: Probably the best idea, yes.

sushain97:
> The module persistence.js is making trouble.
> It overrides some variables imported from other modules.

That means it treats them as globals and can **modify** them. `persistence.js` retrieves values from cache to hydrate the interface.

Androbin:
So, what shall we do about this?
```
Error: assets/js/persistence.js:256
256:             locale = store.get('locale', '');
                 ^^^^^^ locale. import cannot be reassigned
274: /*:: import {iso639Codes, iso639CodesInverse, locale} from "./localization.js" */
                                                   ^^^^^^ import locale
```

sushain97:
> So, what shall we do about this?

I guess we'll just have to either disable flow on the file or suppress the warnings.

sushain97:
Or... maybe you could add a function in the appropriate files that takes an input and just sets the appropriate variable equal to its input... seems somewhat ugly though.

Androbin:
After updating the Type definitions for jQuery, I got this:
```
Library parse error:
interfaces/jquery.js.flow:37
 37:     beforeSend? (jqXHR: JQueryXHR, settings: JQueryAjaxSettings): any;
                     ^ Unexpected token (


Found 1 error
```
[This line](https://github.com/goavki/apertium-html-tools/pull/229/files#diff-ab5e765b62527a795906239d13273103R37) and others seem to use a new syntax.

Androbin:
`beforeSend?: (jqXHR: JQueryXHR, settings: JQueryAjaxSettings) => any;`
became
`beforeSend? (jqXHR: JQueryXHR, settings: JQueryAjaxSettings): any;`

Androbin:
> Large problem is that by adding proper jQuery typings, a ton of errors popped up (50+) and now they need to be corrected before any of this is mergeable.

These errors existed before Flow discovered them. But it would be nice to get the `3.2.1` bindings to work which should get rid of so many nasty event-related errors.

Androbin:
```
Library parse error:
interfaces/jquery.js.flow:28
 28:     export = jQuery;
         ^^^^^^ Unexpected token export

Found 1 error
```
@sushain97 What shall we do about [this line](https://github.com/goavki/apertium-html-tools/pull/229/commits/86491b6c0f7bdb12fe8f826fca9385e5a9090dc8#diff-ab5e765b62527a795906239d13273103R32)?

Androbin:
Seems like the new type definitions are valid TypeScript but not valid Flow:
https://github.com/babel/babel/issues/1151#issuecomment-89411509

Androbin:
* Most reported errors are related to some type definitions being out-of-date.
* Updating is cumbersome because the ones available are valid TypeScript but not valid Flow.
* Apart from that, coverage is now 85% (was 54%)
* CircleCI now runs "flow-coverage-report ..."
* Errors in "flow check" are suppressed for now

assets/js/translator.js
Androbin: Couldn't find a more "idiomatic" way to do it. This is kinda ugly.

sushain97: That's alright - not a big deal.

Androbin:
I have trouble fixing the remaining 44 errors.

* It goes nuts on the signatures in `compat.js`
* I'm not sure where to put stubs for `jsonp`
* @sushain97 broke the import of `config.js` by moving it

There are many issues due to incompatible types:
* lots due to possibly uninitialized variables
* string array literals vs. mutable string arrays

Event Listeners are kinda flawed:
* Additional parameters need to be declared aren't used which triggers other linters
* Types need to be casted one way or the other

@sushain97 might wanna take a quick look at the error messages.

Androbin:
* added a stub for `jsonp`, errors gone
* figured out a few jQuery calls, 30 errors left

Androbin:
Three major issues:
* Possibly uninitialized values in `persistence.js` imported from `translator.js`
* Incompatible signatures in `compat.js`
* Module `config.js` not found anymore (see above)

sushain97:
> Incompatible signatures in compat.js

I think we should just revert all changes to this file, not add `@flow` to it and not include it in coverage checking. It's basically all code from Mozilla anyway and I don't think it'll ever properly typecheck.

I'll take a look at the rest of the stuff later today.

Androbin:
There is [one line in `init.js`](https://github.com/goavki/apertium-html-tools/pull/229/files#diff-48d86820c591610eb1c00727bf90865bR97) which doesn't look quite right:
It's calling `persistChoices` from `persistence.js` without arguments.
But I'm pretty sure the argument is non-optional.

Androbin:
The module `store.js` doesn't really benefit from Flow annotations either.

Androbin:
There are only 9 errors left, due to 3 reasons:
* `init.js` calls `persistChoices` from `persistence.js` without the (presumably) required argument.
* `localization.js` has an issue concerning the values covered by `iso639Codes` and `languages`.
* `persistence.js` might in theory use `curSrcLang` or `curDstLang` from `translator.js` uninitialized.

unhammer:
> The module store.js doesn't really benefit from Flow annotations either.

Doesn't it already have Flow annotations? The annotations should help avoiding e.g. `var x/*:int*/ = store.get("key", "2")` (string as fallback for an int)

Androbin:
@unhammer Can you help with `Store.prototype.get` ?
The return type should really be `?T` but that would break everything.
For now, I made the return type `T|any` which effectively silences the errors.

Androbin:
I was not quite sure about the original intent behind [this line in `translator.js`](https://github.com/goavki/apertium-html-tools/pull/229/files#diff-8331c23a07e5eb815c5b4393e81715efR1009).
But I did my best to infer it from the data types involved.

Androbin:
@sushain97 Can you comment on the last few comments?

assets/js/analyzer.js
```diff
@@ -56,7 +61,7 @@ function getAnalyzers() {
             deferred.resolve();
         }
         else {
-            console.warn('Analyzers cache ' + (analyzers === null ? 'stale' : 'miss') + ', retrieving from server');
+            console.warn('Analyzers cache ' + (analyzers ? 'miss' : 'stale') + ', retrieving from server');
```
sushain97: Seems backwards.

assets/js/analyzer.js
```diff
@@ -23,9 +26,11 @@ if(modeEnabled('analyzation')) {
             persistChoices('analyzer');
         });
 
-        $('#morphAnalyzerInput').keydown(function (e) {
-            if(e.keyCode === ENTER_KEY_CODE && !e.shiftKey) {
-                e.preventDefault();
+        $('#morphAnalyzerInput').keydown(function (ev /*: JQueryKeyEventObject */) {
+            var event /*: JQueryKeyEventObject */ = (ev /*: any */);
```
sushain97: This extra var isn't required for me locally.

Makefile
```diff
@@ -91,10 +91,10 @@ build/js/compat.js: assets/js/compat.js build/js/.d
 	cp $< $@
 
 build/js/jquery.min.js: build/js/.d
-	curl -Ss 'http://ajax.googleapis.com/ajax/libs/jquery/1.11.3/jquery.min.js' -o $@
```
sushain97: Seems like these changes should be reverted since the bindings are for 1.x now.

unhammer:
> @unhammer Can you help with `Store.prototype.get` ?
> The return type should really be `?T` but that would break everything.
> For now, I made the return type `T|any` which effectively silences the errors.

The return type should be `T` (it should always be called with a fallback of the same type as the expected return type). What's the error? Flow used to be fine with it …

Androbin:
@unhammer Okay, but what about [this line](https://github.com/goavki/apertium-html-tools/blob/master/assets/js/store.js#L27)?

Androbin:
```
Error: assets/js/store.js:29
 29:         return undefined;
                    ^^^^^^^^^ undefined. This type is incompatible with the expected return type of
 15: Store.prototype.get = function /*:: <T> */ (key /*: string */, fallback /*: T */) /*: T */ {
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ some incompatible instantiation of `T`
```

assets/js/analyzer.js
Androbin: The original ternary was true for null values.
The new ternary is true for non-null values.
Also, I flipped the respective values.
Should be fine.

Androbin:
Resubmitted as #234

unhammer:
Robin Richtsfeld <notifications@github.com> čálii:

> @unhammer Okay, but what about [this
> line](https://github.com/goavki/apertium-html-tools/blob/master/assets/js/store.js#L27)?

Oh, that looks like a bug. It should return fallback there.

Androbin:
> Oh, that looks like a bug. It should return fallback there.

Thought so, too. [Already changed it](https://github.com/goavki/apertium-html-tools/pull/234/files#diff-0aba10d24fae729bf34c4c4a9fae7516R24).

PR #230, opened 2017-12-19
Issue 229 - flow coverage

sushain97:

PR #234, opened 2017-12-19
Better flow coverage

Androbin:
@sushain97 Friendly ping.


Androbin:
I've just noticed that `config.js.flow` declares some constants as non-null that we don't have fallbacks for in `read-conf.py`, namely:
`ALLOWED_LANGS`, `ALLOWED_PAIRS`, `ALLOWED_VARIANTS`, `SUBTITLE`, `SUBTITLE_COLOR`

sushain97:
@Androbin good point. I fixed the last of them: https://github.com/goavki/apertium-html-tools/pull/230/commits/31dab4a3ccbb562d2c69b4d9842df3a72b452da5

PR #239, opened 2017-12-25
Fix buggy UI with detect-language.

### LINKED ISSUE #236: Buggy UI after switching into translate document/webpage after detecting language
No language is selected anymore:

image

JPJPJPOPOP:
Fixes #236.

### DIFF
--- a/assets/js/translator.js
+++ b/assets/js/translator.js
@@ -89,6 +89,19 @@ if(modeEnabled('translation')) {
                     $('div#translateText').fadeIn('fast', function () {
                         synchronizeTextareaHeights();
                     });
+                    if($('#detect').hasClass('activeAfterCancel')) {
+                        $('.srcLang').removeClass('active');
+                        $('#detect').addClass('active');
+                        muteLanguages();
+                        $('#detectedText').parent('.srcLang').attr('data-code', curSrcLang);
+                        refreshLangList();
+                        $('#detectedText').show();
+                        $('#detectText').hide();
+                        $('#detect').removeClass('activeAfterCancel');
+                    }
+                    else {
+                        translateText();
+                    }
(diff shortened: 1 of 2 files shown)


shardulc:
@sushain97 Works perfectly fine for me. If selecting `srcLang1` by default is OK (when would it not be?), I think this can be merged.

assets/js/translator.js
```diff
@@ -89,6 +89,13 @@ if(modeEnabled('translation')) {
                     $('div#translateText').fadeIn('fast', function () {
                         synchronizeTextareaHeights();
                     });
+                    if($('#detect').hasClass('originDetect') === true) {
+                        $('#srcLang1').removeClass('active');
```
sushain97: I think `.srcLang` would work here and below.

JPJPJPOPOP: @sushain97 ok, changing.

assets/js/translator.js
```diff
@@ -111,6 +116,11 @@ if(modeEnabled('translation')) {
                     $('#fileInput').show();
                     $('div#fileName').hide();
                     $('div#docTranslation').fadeIn('fast');
+                    if($('#detect').hasClass('active') === true) {
+                        $('#srcLang1').addClass('active');
```
sushain97: I’m somewhat afraid of doing things in a more ad hoc manner like this because just changing the classes around might not change some of the program state that should also have been changed.

`$('#srcLang1').click()` might be better? and when switching back, do some of the things in `handleDetectLanguageSuccessResponse` if detect was selected?

JPJPJPOPOP: @sushain97
When you say "do some of the things in `handleDetectLanguageSuccessResponse` if detect was selected", do you mean move the stuff that is after $('#detect').hasClass('active') === true?

PR #242, opened 2017-12-27
Use activeAfterCancel for 'Detect Language' consistently

### LINKED ISSUE #241: Detect Language fails for empty text
The interface doesn't catch attempts to detect the language of empty text which results in https://www.apertium.orf/apy/identifyLang producing code 400 errors.

screenshot

Androbin:
Fixes #241 
@sushain97 Friendly ping.

### DIFF
--- a/assets/js/translator.js
+++ b/assets/js/translator.js
@@ -58,10 +58,10 @@ if(modeEnabled('translation')) {
             });
 
             $('#originalText').on('input propertychange', function () {
-                updateDetect($('#originalText').text() !== '');
+                updateDetect($('#originalText').val() !== '');
             });
 
-            updateDetect($('#originalText').text() !== '');
+            updateDetect($('#originalText').val() !== '');
         }
 
         function setupWebpageTranslation() {
--- a/assets/js/translator.js
+++ b/assets/js/translator.js
@@ -994,13 +994,13 @@ function showTranslateWebpageInterface(url /*: ?string */, ignoreIfEmpty /*: ?bo
     });
 }
 
-function updateDetect(state /*: bool */) {
-    var oldState;
+function updateDetect(active /*: bool */) {
+    var wasActive;
 
-    if(state) {
-        oldState = $('#detect').hasClass('activeAfterCancel');
+    if(active) {
+        wasActive = $('#detect').hasClass('activeAfterCancel');
 
-        if(oldState) {
+        if(wasActive) {
             $('.srcLang').removeClass('active');
             $('#detect').addClass('active');
             handleDetectLanguageSuccessComplete();
@@ -1011,9 +1011,9 @@ function updateDetect(state /*: bool */) {
         $('#detect').removeClass('disabledLang');
     }
     else {
-        oldState = $('#detect').hasClass('active');
+        wasActive = $('#detect').hasClass('active');
 
-        if(oldState) {
+        if(wasActive) {
             $('#srcLang1').click();
             $('#detect').addClass('activeAfterCancel');
         }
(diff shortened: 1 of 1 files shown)


sushain97:
screenshots? it seems like your code hides instead of disabling.

check out what happens when I switch to website translation -- it should be similar

Androbin:
Just noticed, this logic was already implemented but not triggered on load.

sushain97:
Code looks alright. How does this mesh with the changes in https://github.com/goavki/apertium-html-tools/pull/239? Does it get updated correctly?

Androbin:
We should probably make use of the `activeAfterCancel` class introduced.

Androbin:
@sushain97 Are you okay with the lastest changes?

assets/js/translator.js
```diff
@@ -29,6 +29,35 @@ var PUNCTUATION_KEY_CODES = [46, 33, 58, 63, 47, 45, 190, 171, 49]; // eslint-di
     callApy, apyRequestTimeout, isURL */
 /* global ENTER_KEY_CODE, HTTP_BAD_REQUEST_CODE, HTTP_OK_CODE, SPACE_KEY_CODE, XHR_DONE, XHR_LOADING */
 
+function updateDetect(state /*: bool */) {
```
sushain97: Could you move this function to a more appropriate location?

Androbin: In terms of module or what?

Androbin: What would be the alternative?

sushain97: Somewhere with the other functions? i.e. around L1026

Androbin:
@sushain97 Are you satisfied with the PR?

assets/js/translator.js
```diff
@@ -58,12 +58,10 @@ if(modeEnabled('translation')) {
             });
 
             $('#originalText').on('input propertychange', function () {
-                var disableDetect = this.value === '';
-                $('#detect, #srcLangSelect option[value="detect"]').prop('disabled', disableDetect);
-                $('#detect').toggleClass('disabledLang', disableDetect);
-
-                persistChoices('translator');
+                updateDetect($('#originalText').text() !== '');
```
sushain97: What would happen if this triggers with `false` here and then `true`? There's no `true` call corresponding to your `$('#detect, #srcLangSelect option[value="detect"]').prop('disabled', true);`.

Androbin: Interesting, none of the previous code did this properly.

Androbin:
Okay, I know what you meant. Should be more logical now.

assets/js/translator.js
```diff
@@ -1018,6 +994,36 @@ function showTranslateWebpageInterface(url /*: ?string */, ignoreIfEmpty /*: ?bo
...
 
+function updateDetect(state /*: bool */) {
+    var oldState;
+
+    if(state) {
+        oldState = $('#detect').hasClass('activeAfterCancel');
+
+        if(oldState) {
+            $('#detect').removeClass('activeAfterCancel');
+            $('#detect').click();
```
sushain97: This is problematic. Consider the following sequence of events:

1. I type some text in
2. I click detect (it detects Spanish)
3. I click website translation (it switches to Spanish in the recent languages NOT detect and disables the detect button)
4. I cancel website translation and go back to text translation

The current code would click the detect button and a new request would be sent to APy. The old code doesn't do that since we already know what the detected language is and don't need a new request, hence the call to `detectLanguageSuccess...` only.

Androbin: Okay, not I get it. Thanks for the explanation.

Androbin:
Does this look okay now?

Androbin:
How/Where does the input field get populated from the query string?

Androbin:
Seems like the `inputchange` wasn't triggered...

Androbin:
Okay, got it. I was testing `.text()` which is of course rubbish.

Issue #130, opened 2017-04-12
faster, slimmer, lighter


unhammer:
Firefox says we download 519KB on an empty cache:

* 7 js files totalling 279,99KB
* 4 css files totalling 179,29KB
* 1 html file totalling 40,47KB
* 4 images totalling 18,72KB

It's not too bad (half the size of google.com!), but still about as large [as one of the lesser Russian classics](http://idlewords.com/talks/website_obesity.htm). And we are after all just serving text.

Possible ideas:
* Is it possible to trim bootstrap down to only what we use? Or exchange it for something both smaller and nicer?
* Should we try using something smaller-but-equivalent like [jquip](https://github.com/mythz/jquip) instead of jquery? 
* Are there parts of our own js that is not used, that we can cut out?

More ideas:
http://www.lightentheweb.com/

share-with-me:
> Is it possible to trim bootstrap down to only what we use? Or exchange it for something both smaller and nicer?

> Are there parts of our own js that is not used, that we can cut out?

IMO, we could start with these 2 ideas. We can make use of JS code coverage tool for instance,

share-with-me:
> Is it possible to trim bootstrap down to only what we use? Or exchange it for something both smaller and nicer?

> Are there parts of our own js that is not used, that we can cut out?

We can definitely start with these. A JS code coverage tool such as [JSCover](http://tntim96.github.io/JSCover/) is a tool that displays a summary of the code coverage and an example for the same is [here](http://tntim96.github.io/JSCover/example-report-yui3/jscoverage.html?test/index.html).

sushain97:
There are existing tools out there that are excellent at tree shaking. However, it would complicate the build process and require extra dependencies: https://webpack.js.org/guides/tree-shaking/.

Personally, I would be all for switching to a more context appropriate build process :) We could always create a branch/releases that have a compiled ready-to-execute version of the site.

unhammer:
Hm, when I said "cut out" I wasn't thinking of doing this at compile time, but rather actually cutting out the code that we don't use from the *source* – dead code belongs in branches :)

sushain97:
I agree that dead code belongs in branches. However, I haven't checked
carefully but I would venture that very little of our JS is dead. ESLint
errors on lots of potential causes and I'm usually relatively careful with
it. I think the majority of what we serve is e.g. jQuery, that we couldn't
really cut down manually.

Manually trimming Bootstrap would make it really tricky to upgrade
[...]

PR #243, opened 2017-12-30
Minify all.js and all.css

Androbin:
Towards #130


sushain97:
Hm, Python libraries aren't exactly light in terms of dependencies either. However, I think this is an OK solution if and only if an individual without the libraries attempts to compile the app, it still works, it's just not minified. If you could make those changes and modify the README this looks good to me.

Androbin:
If a dependency is not installed, minification will just skip and print a hint.
I have updated the README section on compression accordingly.

sushain97:
I believe at least `csscompressor` has a built in command line version that would be preferable.

Also, please modify the Docker configs to also install the new dependencies.

Androbin:
Okay, I have updated the `Makefile` and `Dockerfile` accordingly.

sushain97:
This seems to fully break the app:

image

image

sushain97:
Also, please don't force push. I made some commits that you should pull down that got overwritten last time.

README.md
```diff
@@ -44,8 +44,8 @@ Any changes to the local version of Html-tools will trigger a `make`.
 
 For production usage, remember to
 
-1. Build with `jsmin` and `csscompressor` installed through `pip`
-2. Enable gzip compression on your server
+1. Build with `jsmin` and `csscompressor` installed through `pip`.
+1. Enable gzip compression on your server.
```
Androbin: Why?

sushain97: We've had this conversation before. It makes diffs smaller if you need to insert something.

Androbin:
@sushain97 I can't reproduce that error on my machine, neither with nor without `jsmin` installed. I have even put that file through multiple syntax checkers. Could you check if it's a problem on your side?

sushain97:
@Androbin can you post updated versions of the numbers on #130?

Androbin:
Do the minified versions already get shipped?

PR #245, opened 2018-01-01
Ask for less font(s)

Androbin:
Towards #130 
@sushain97 FYI


sushain97:
Please see my comment on #130. We don't only use the Latin script.

Issue #82, opened 2016-12-18
Language variant picker


sushain97:
Displaying language variants as distinct languages in the translator language selector is awkward and repetitive. Allowing users to first select a language and then display radio buttons for choosing a variant below the relevant translation box, if relevant, provides a better user interface.

The variants would be sorted alphabetically and the first valid option would be selected. Options would continue to be persisted and variant preferences would be stored. If there is only one variant available for a given language, it would be displayed and checked by default.

The proposed implementation would require modification for mobile users since both the language selector and radio buttons would interfere with the flow of the page and not both fit on the same viewport, likely causing confusion. A possible alternative is including the variants as a regular language in the language dropdown for mobile.

unhammer:
How would the workflow be for translating between the variants? Like cat→cat_valencia or whatever

sushain97:
You would select Catalan on both sides and it would default to using a variant on the destination side != the target side. Sound good?

unhammer:
My worry is that the UI turns a bit "undiscoverable" – I certainly wouldn't guess that Apertium could translate into Valencian if I didn't see it in the language list on my first click, so how would I know that I had to select Catalan on both sides to do that? 

One possible solution would be to somehow mark languages-with-variants differently in the (target) language selector, but I can't think of a good way that's both informative and short …

sushain97:
We could have the language selector have a dropdown beside the language name (to indicate variants) but then it's a dropdown within a dropdown...

unhammer:
Hm, what if we do like this:

![variants](https://cloud.githubusercontent.com/assets/56868/21452657/bdc28220-c908-11e6-85cd-b330b3ef6bca.png)

and then when someone clicks "katalansk", it defaults to "cat", but that second selector shows?

ryanachi:
I'm planning on working on this issue next.  Do you think we'll be going ahead with @unhammer 's proposal, or we planning to integrate @sushain97 's 'dropdown-within-dropdown' idea?

sushain97:
@shardulc also has a proposal that we discussed in IRC that could be good. perhaps he can copy the discussion here

shardulc:
On [2017-12-26 on #apertium](https://tinodidriksen.com/pisg/freenode/logs/%23apertium_20171226.log):

```
[05:53:13]
<sushain> we really need to have better variant support :(
<sushain> we’ve needed it for like 4 years
<shardulc> variant = eng_US, cat_valencia, etc. right?
[...]

PR #256, opened 2018-01-06
Improved interface for language variants

ryanachi:
This PR addresses https://github.com/goavki/apertium-html-tools/issues/82 (language variants picker). I've modified:

/assets/js/translator.js, to sort variants correctly
/assets/css/translation.css, to provide proper styling for variants
I've also added a new SQL file for variants in goavki/apertium-apy#78.

Please let me know if you have any comments!

### DIFF
--- a/assets/css/translation.css
+++ b/assets/css/translation.css
@@ -7,12 +7,12 @@ html[dir='rtl'] .languageName {
 }
 
 html[dir='ltr'] .languageVariant {
-	font-size: 12px;
+    font-size: 12px;
     padding-left: 2.5em;
 }
 
 html[dir='rtl'] .languageVariant {
-	font-size: 12px;
+    font-size: 12px;
     padding-right: 2.5em;
 }
 
--- a/assets/css/translation.css
+++ b/assets/css/translation.css
@@ -6,6 +6,16 @@ html[dir='rtl'] .languageName {
     padding-right: .5em;
 }
 
+html[dir='ltr'] .languageVariant {
+	font-size: 12px;
+    padding-left: 2.5em;
+}
+
+html[dir='rtl'] .languageVariant {
+	font-size: 12px;
+    padding-right: 2.5em;
+}
+
 .languageName {
     position: relative;
 }
--- a/assets/js/translator.js
+++ b/assets/js/translator.js
@@ -533,6 +533,9 @@ function populateTranslationList() {
     sortTranslationList();
     $('.languageCol').remove();
 
+function isVariant(lang) {
+return lang.indexOf('_') !== -1;
+}
     var minColumnWidth = TRANSLATION_LIST_MAX_WIDTH / TRANSLATION_LIST_MAX_COLUMNS;
 
     // figure out how much space is actually available for the columns
@@ -560,13 +563,13 @@ function populateTranslationList() {
         dstLangsPerCol = Math.ceil(dstLangs.length / numDstCols);
 
     for(var i = 0; i < numSrcCols; i++) {
-        while(srcLangs[i * srcLangsPerCol].indexOf('_') !== -1) {
+        while(isVariant(srcLangs[i * srcLangsPerCol])) {
             srcLangsPerCol++;
         }
     }
 
     for(i = 0; i < numDstCols; i++) {
-        while(dstLangs[i * dstLangsPerCol].indexOf('_') !== -1) {
+        while(isVariant(dstLangs[i * dstLangsPerCol])) {
             dstLangsPerCol++;
         }
     }
@@ -580,7 +583,7 @@ function populateTranslationList() {
                 var langCode = srcLangs[j];
                 var langName = getLangByCode(langCode);
                 var langClasses = 'languageName';
-                if(typeof langCode !== 'undefined' && langCode.indexOf('_') !== -1) {
+                if(typeof langCode !== 'undefined' && isVariant(langCode)) {
                     langClasses += ' languageVariant';
                 }
                 srcLangCol.append(
@@ -601,7 +604,7 @@ function populateTranslationList() {
                 langCode = dstLangs[j];
                 langName = getLangByCode(langCode);
                 langClasses = 'languageName';
-                if(typeof langCode !== 'undefined' && langCode.indexOf('_') !== -1) {
+                if(typeof langCode !== 'undefined' && isVariant(langCode)) {
                     langClasses += ' languageVariant';
                 }
                 dstLangCol.append(
@@ -621,14 +624,14 @@ function populateTranslationList() {
     $('.langSelect option[value!=detect]').remove();
     $.each(srcLangs, function () {
         var indent = '';
-        if(typeof this !== 'undefined' && this.indexOf('_') !== -1) {
+        if(typeof this !== 'undefined' && isVariant(this)) {
             indent += '&nbsp;&nbsp;&nbsp;&nbsp;';
         }
         $('#srcLangSelect').append($('<option></option>').prop('value', this).html(indent + getLangByCode(this)));
     });
     $.each(dstLangs, function () {
         var indent = '';
-        if(typeof this !== 'undefined' && this.indexOf('_') !== -1) {
+        if(typeof this !== 'undefined' && isVariant(this)) {
             indent += '&nbsp;&nbsp;&nbsp;&nbsp;';
         }
         $('#dstLangSelect').append($('<option></option>').prop('value', this).html(indent + getLangByCode(this)));
@@ -680,12 +683,8 @@ function populateTranslationList() {
                 function containsParentLang(variant) {
                     return variant.indexOf(langVariant[0]) !== -1;
                 }
-                if(pairs[curSrcLang] && (pairs[curSrcLang].indexOf(lang) !== -1 || pairs[curSrcLang].indexOf(langVariant[0]) !== -1 || pairs[curSrcLang].some(containsParentLang))) {
-                    return true;
-				}
-                else {
-                    return false;
-                }
+                return pairs[curSrcLang] && (pairs[curSrcLang].indexOf(lang) !== -1 ||
+                    pairs[curSrcLang].indexOf(langVariant[0]) !== -1 || pairs[curSrcLang].some(containsParentLang));
             }
             var aPossible = isPossible(a), bPossible = isPossible(b);
             if(aPossible === bPossible) {
--- a/assets/js/translator.js
+++ b/assets/js/translator.js
@@ -562,17 +562,16 @@ function populateTranslationList() {
     for(var i = 0; i < numSrcCols; i++) {
         while(srcLangs[i * srcLangsPerCol].indexOf('_') !== -1) {
             srcLangsPerCol++;
-            break;
         }
     }
 
-    for(var i = 0; i < numDstCols; i++) {
+    for(i = 0; i < numDstCols; i++) {
         while(dstLangs[i * dstLangsPerCol].indexOf('_') !== -1) {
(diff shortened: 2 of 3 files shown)


assets/js/translator.js
```diff
@@ -646,8 +653,41 @@ function populateTranslationList() {
...
+            if(srcLangs[i].indexOf('_') !== -1) {
+                variantsOnly.push(srcLangs[i]);
+            }
+            else {
+                langsOnly.push(srcLangs[i]);
+            }
+        }
+
+        srcLangs = langsOnly.sort(compareLangCodes);
+        for(i = 0; i < variantsOnly.length; i++) {
```
shardulc: While the double for-loop works, I think there's a more efficient way to do it:
1. sort `langsOnly`
2. sort `variantsOnly`
3. have `var langHead = 0` and `var variantHead = 0`
4. in a single for-loop, compare the lang at `langsOnly[langHead]` with the variant at `variantsOnly[variantHead]`, and depending on the comparison, either `langHead++` or `variantHead++`; simultaneously build the entire `srcLangs` list

(If that's not clear, ask on IRC :smiley:)

assets/js/translator.js
```diff
@@ -1104,6 +1154,11 @@ function muteLanguages() {
             $(this).addClass('text-muted');
         }
     });
+    $.each($('#dstLanguages .languageVariant'), function () {
```
shardulc: Elements with the `languageVariant` class also have the `languageName` class, so the lines just above this already deal with them, right?

shardulc:
Although it is displayed in the dropdown as just "Valencian", I think in the top bar it should be "Catalan Valencian":

![valencian][https://user-images.githubusercontent.com/10349624/34636584-a58483c2-f259-11e7-9b32-9a1ef08a4f01.png]

@sushain97 Do you agree, or is "Valencian" good?

shardulc:
Another thing I'm not sure about: How does translating between variants work, e.g. Catalan -> Catalan Valencian? As far as I can see, it's not a 'pair' as such in Apertium, and apertium.org doesn't support it, so is it even possible?

ryanachi:
@sushain97 would you like the top bar to display only the variant name, or both the language and variant names?

sushain97:
> Although it is displayed in the dropdown as just "Valencian", I think in the top bar it should be "Catalan Valencian":

Consider for example the following three variants:

- por_BR
- cat_valencia
- eng_US

The language names DB could contain the "fully qualified" versions of each of these:

- Brazilian Portuguese
- Valencian
- American English

or the less fully qualified version:

- Brazilian
- Valencian
- American

Now our display will include the 'parent' language above the variants anyway so having the less qualified versions in the DB is fine UNTIL we try to put them into the recent languages bar when one gets selected. "American" doesn't seem like a good fit there. Consider that "I speak American" sounds somewhat awful. We could just append them together, "American English" is good. But then, "Valencian Catalan" sounds bad (I think?). We could put the variant in parentheses, e.g. "Catalan (Valencian)" or "Portuguese (Brazilian)". While that's less ideal, it does seem to work. Until you consider that we have to support all languages and parentheses are unlikely to be appropriate in all languages (the appending 'solution' probably has the same problem).

So, the solution seems to be either just using the fully qualified versions everywhere or having the DB store both (maybe semicolon delimited? e.g. the following and APy would split them:

- "Brazilian Portuguese;Brazilian"
- "Valencian;Valencian"
- "American;American English"

Storing both like this seems hacky and I think it would make the code pretty complicated. So, my thoughts are we go with the fully qualified version everywhere. Thoughts?

cc @jonorthwash @unhammer @ftyers thoughts on the right approach? have I missed any "cases"?

assets/css/translation.css
```diff
@@ -6,6 +6,16 @@ html[dir='rtl'] .languageName {
     padding-right: .5em;
 }
 
+html[dir='ltr'] .languageVariant {
+	font-size: 12px;
```
sushain97: Indentation?

assets/css/translation.css
```diff
@@ -6,6 +6,16 @@ html[dir='rtl'] .languageName {
...
+    padding-left: 2.5em;
+}
+
+html[dir='rtl'] .languageVariant {
+	font-size: 12px;
```
sushain97: As above

assets/js/translator.js
```diff
@@ -567,8 +567,12 @@ function populateTranslationList() {
             if(numSrcLang < srcLangs.length) {
                 var langCode = srcLangs[j];
                 var langName = getLangByCode(langCode);
+                var langType = 'languageName';
```
sushain97: `langType` -> `langClasses`?

ryanachi:
Adding the full names seems like a good idea; it definitely makes for a thorough solution.  I'll add the full names to the variants.sql file in APy.  Are there any other changes that we would like implemented?

sushain97:
> Are there any other changes that we would like implemented?

Probably should work? :P See below.

image

ryanachi:
That's odd; I'm getting this when I test. https://imgur.com/a/KP4zm

assets/js/translator.js
```diff
@@ -646,8 +657,72 @@ function populateTranslationList() {
...
+
+        var langHead = 0;
+        var variantHead = 0;
+        var numLangs = srcLangs.length;
+        srcLangs = [];
+
+        langsOnly = langsOnly.sort(compareLangCodes);
+        variantsOnly = variantsOnly.sort(compareLangCodes);
+
+        while((langHead + variantHead) < numLangs) {
```
sushain97: I'm a little confused by this code. If the point of it is to get the order correct, I think `compareLangCodes` might already do that? Also, it should be extracted into a function since its used twice and is pretty long ;)

sushain97: (I didn't read the code super carefully so I could be misunderstand the intent)

jonorthwash:
> Consider that "I speak American" sounds somewhat awful. 

[Not necessarily](http://www.pbs.org/speak/). But yeah, that's not what we want here.

ryanachi:
@sushain97 I've incorporated the three suggestions that you made: https://395-28576409-gh.circle-artifacts.com/0/home/circleci/project/build/index.eng.html?dir=slv-bos#translation

assets/js/translator.js
```diff
@@ -647,20 +682,7 @@ function populateTranslationList() {
...
-                return compareLangCodes(a, b);
-            }
-            else if(aPossible && !bPossible) {
-                return -1;
-            }
-            else {
-                return 1;
-            }
-        });
+        dstLangs = dstLangs.sort(compareLangCodes);
```
sushain97: Why was this deleted? The extra sorting puts available languages first. It might require some modification for variants properly displaying though.

assets/js/translator.js
```diff
@@ -635,10 +670,10 @@ function populateTranslationList() {
             if(aVariant[1] && bVariant[1] && aVariant[0] === bVariant[0]) {
                 return directCompare;
             }
-            else if(aVariant && a.startsWith(b)) {
+            else if(aVariant[1] && aVariant[0] === b) {
```
sushain97: Good catch.

assets/js/translator.js
```diff
@@ -600,10 +624,22 @@ function populateTranslationList() {
...
+                indent += '&nbsp;&nbsp;&nbsp;&nbsp;';
+            }
+        }
+        $('#srcLangSelect').append($('<option></option>').prop('value', this).html(indent + getLangByCode(this)));
     });
     $.each(dstLangs, function () {
-        $('#dstLangSelect').append($('<option></option>').prop('value', this).text(getLangByCode(this)));
+        var indent = '';
+        if(typeof this !== 'undefined') {
+            if(this.indexOf('_') !== -1) {
```
sushain97: Can these be combined into a single if with a conjunction?

ryanachi:
Done! https://397-28576409-gh.circle-artifacts.com/0/home/circleci/project/build/index.eng.html?dir=slv-bos#translation

sushain97:
Now:

image

Before:

image

It doesn't look like we're getting the same behavior where available languages are sorted first.

sushain97:
sad:

image

ryanachi:
@sushain97 I've revised the sorting function to correctly display the language variants.

assets/js/translator.js
```diff
@@ -661,6 +699,16 @@ function populateTranslationList() {
                 return 1;
             }
         });
+
+        for(i = 0; i < variantsOnly.length; i++) {
```
sushain97: Just to confirm - this works when a variant is possible but the parent language isn't and vice versa, right? Could you provide screenshots for each case? Also, this nested for loop seems a bit expensive?

ryanachi: Yes, that's correct—I've attached a screenshot of a case when the parent language is available but the variant language isn't.  As for the converse scenario, apertium.org doesn't have a pair where the variant language is available but the parent language isn't.  However, it should work as well, as the 'for' loop inserts each variant language immediately after its respective parent language, no matter what the position is—this can be seen in the screenshot.

I agree, the 'for' loop is a bit inefficient; however, I couldn't find a better way to sort the list of languages so that each variant appears immediately after its parent language.  I'd be happy to implement another solution if we can find one. :)

sushain97:
What if you sorted the same way (built in sort) but considered a variant
available if it's actually available, it's parent is available or another
variant with the same parent is available. And for a non-variant, available
if it's actually available or any of it's variants is available. Building a
hash from language to its variants will make this effectively linear,
right? Splicing is also pretty expensive when you do it over and over again.

sushain97:
Also, try it out with beta.apertium.org/apy, it's probably easier to find a
variant that's possible with a parent that isn't then.

On Jan 9, 2018 12:47 PM, "Sushain Cherivirala" <sushain@skc.name> wrote:

What if you sorted the same way (built in sort) but considered a variant
available if it's actually available, it's parent is available or another
variant with the same parent is available. And for a non-variant, available
if it's actually available or any of it's variants is available. Building a
hash from language to its variants will make this effectively linear,
right? Splicing is also pretty expensive when you do it over and over again.

assets/js/translator.js
```diff
@@ -648,9 +675,22 @@ function populateTranslationList() {
 
         srcLangs = srcLangs.sort(compareLangCodes);
         dstLangs = dstLangs.sort(function (a, b) {
-            var aPossible = pairs[curSrcLang] && pairs[curSrcLang].indexOf(a) !== -1;
-            var bPossible = pairs[curSrcLang] && pairs[curSrcLang].indexOf(b) !== -1;
-
+            function determineIfPossible(a) {
```
sushain97: Use `isPossible(x)` instead so we don't shadow `a`.

assets/js/translator.js
```diff
@@ -648,9 +675,22 @@ function populateTranslationList() {
...
+                var aVariant = a.split('_');
+                if(pairs[curSrcLang] && pairs[curSrcLang].indexOf(a) !== -1) {
```
sushain97: Also this should just be a huge return with each of these things as parenthesized lines in a conjunction :)

assets/js/translator.js
```diff
@@ -648,9 +675,22 @@ function populateTranslationList() {
...
+                    return true;
+                }
+                else if(pairs[curSrcLang] && pairs[curSrcLang].indexOf(aVariant[0]) !== -1) {
+                    return true;
+                }
+                else if(pairs[curSrcLang] && pairs[curSrcLang].join().match('/' + aVariant[0] + '/g')) {
```
sushain97: This is really hacky..

sushain97: Maybe `.filter`?

PR #271, opened 2018-01-21
Compress HTML

Androbin:
Towards #130
#243 @sushain97 /cc

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -134,7 +134,9 @@ build/l10n-rel.html: assets/strings/locales.json isobork build/.d
 	awk 'BEGIN{while(getline<"isobork")i[$$1]=$$2} /:/{sub(/^[^"]*"/,""); sub(/".*/,""); borkd=i[$$0]; if(!borkd)borkd=$$0; print "<link rel=\"alternate\" hreflang=\""borkd"\" href=\"index."$$0".html\">"}' $^ > $@
 
 build/index.%.html: build/strings/%.json build/index.localiseme.html $(CONFIG) tools/read-conf.py tools/localise-html.py
-	./tools/localise-html.py -c $(CONFIG) build/index.localiseme.html $< $@
+	./tools/localise-html.py -c $(CONFIG) build/index.localiseme.html $< $@.tmp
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -134,7 +134,9 @@ build/l10n-rel.html: assets/strings/locales.json isobork build/.d
 	awk 'BEGIN{while(getline<"isobork")i[$$1]=$$2} /:/{sub(/^[^"]*"/,""); sub(/".*/,""); borkd=i[$$0]; if(!borkd)borkd=$$0; print "<link rel=\"alternate\" hreflang=\""borkd"\" href=\"index."$$0".html\">"}' $^ > $@
 
 build/index.%.html: build/strings/%.json build/index.localiseme.html $(CONFIG) tools/read-conf.py tools/localise-html.py
-	./tools/localise-html.py -c $(CONFIG) build/index.localiseme.html $< $@
+	./tools/localise-html.py -c $(CONFIG) build/index.localiseme.html $< $@.tmp
```
sushain97: Seems possible to just pipe into `htmlmin`.

Androbin: The script expects an output path. It doesn't print to `stdout`.

sushain97: https://htmlmin.readthedocs.io/en/latest/quickstart.html#command-line

Androbin: I was referring to `localise-html.py` not printing to `stdout`.
I am well aware that I could pipe that into `htmlmin` if it did.

sushain97: Change it then?

Androbin:
* `tools/localise-html.py` now prints to `stdout`
* `Makefile` adapted accordingly

sushain97:
How will this work if `htmlmin` isn't available?

Androbin:
I can provide a fallback like last time. But then we'd better return to writing the input html to a file first.

Androbin:
Should work now.

Androbin:
I was not quite sure about the other HTML targets.
Are these just intermediate build steps or do we need to do this for other files as well?

Androbin:
@sushain97 Anything else to consider?

PR #289, opened 2018-03-09
Update Aragonese translation

Juanpabl:
I updated the Aragonese translation


sushain97:
Thanks, @Juanpabl ! Seems like this could use a small JSON syntax fix? (hence build fail)

Juanpabl:
Sorry, now it's ok

sushain97:
Great! This is mergeable after the two commands in the README are run:

https://github.com/apertium/apertium-html-tools/tree/master/assets/strings#localization

I can take care of it tonight and then merge if you'd like.

Juanpabl:
I've tried to do it myself but it seems that I have something misconfigured and does not work, so please do it for me. Thanks!

sushain97:
done in #290

PR #294, opened 2018-03-16
Remove unused CSS

Androbin:


Androbin:
@sushain97 I don't know about our long-term solution.
But this should make `purifycss` superfluous for now...

sushain97:
Like I said on the other issue, I don't see manually editing the CSS as a viable solution.

Androbin:
Like I said, not long-term. But those unused lines just pile up. Might as well just delete them now.

sushain97:
No... what if I want functionality of Bootstrap that you deleted? Or I want to upgrade Bootstrap? Both those actions would be a pain.

Androbin:
Maybe we'll find an alternative to `purifycss`...

Issue #114, opened 2017-03-13
Get rid of inline styles


sushain97:
There are 38 inline styles in `index.html.in`. They should be eliminated to enable to the `inline-style-disabled` HTMLHint rule.

share-with-me:
Created a PR #117 . Please have a look @sushain97

idealadarsh:
@sushain97 I'd like to work on this issue.

shardulc:
@sushain97 Fix with #214, should be closed.

sushain97:
Not yet, unfortunately!

PR #295, opened 2018-03-20
Removing inline styles in index.html.in #114

SNikhilVashista:
Replaced all the style tags in **index.html.in** with respective IDs  and the IDs are added into a new .css(**removeinline.css**)file in _/apertium-html-tools-master/assets/css_, further, included this **removeinline.css** as a new file in Makefile.

Looking forward to here from you.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -180,6 +180,7 @@ CSSFILES= \
 	assets/css/footer.css \
 	assets/css/general.css \
 	assets/css/navbar.css \
+	assets/css/removeinline.css \
 	assets/css/translation.css
 
 THEMES= cerulean cosmo cyborg darkly journal lumen paper readable sandstone simplex slate spacelab superhero united yeti
--- a/assets/css/removeinline.css
+++ b/assets/css/removeinline.css
@@ -1,6 +1,7 @@
- 
-#one{position: relative;}
-#two { 
+#one {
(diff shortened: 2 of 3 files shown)


share-with-me:
@SNikhilVashista, please ensure that the _id_ names convey the usage of the element being styled. Id names such as #one and so forth are completely random and do not (at all) give any idea about the element being styled.

SNikhilVashista:
Hi Monish, yeah I tried to name them of their usage but I couldn't think of
any as it is for removing inline styling so please suggest if any thought
on name... And by the way the names weren't random they are sequential :p

On 20-Mar-2018 3:46 PM, "Monish Godhia" <notifications@github.com> wrote:

@SNikhilVashista <https://github.com/snikhilvashista>, please ensure that
the *id* names convey the usage of the element being styled. Id names such
as #one and so forth are completely random and do not (at all) give any
idea about the element being styled.

—
<https://github.com/apertium/apertium-html-tools/pull/295#issuecomment-374544288>,
or mute the thread

sushain97:
I suggest you take some inspiration from https://github.com/apertium/apertium-html-tools/pull/214 and the inline-styles branch.

sushain97:
(having everything styled by effectively random ID is just as good as inline)

PR #298, opened 2018-03-24
#114 tried to remove inline styling in index.html.in

aditya-369:
1) deleted  inline styles in index.html.in added id's to required tags.
2)created a removeinline.css file and added style in this file.
3)updated this file in "make" file


sushain97:
The styles shouldn't go in one massive `removeinline.css`. They should go in the most appropriate CSS files that already exist. Also, we should be backporting utility CSS classes from Bootstrap 4 to make the transition easier rather than creating a bunch of repetitive rule sets.

aditya-369:
Hi sushain97,
         I tried to add the styles in style.yeti.css but the styles did not appear in styles.css after running make so i did not know where i was wrong.so i added a file.Please help me in adding the styles in existing files.

JPJPJPOPOP:
@aditya-369 You should be adding them to the css files in `assets/css`. The css have already been divided into multiple files, i.e. `navbar.css`, `translation.css`, etc. There should be no need to be touching any of the Bootstrap themes.

aditya-369:
#300

PR #302, opened 2018-03-25
Updated for circle ci

aditya-369:


assets/css/general.css
```diff
@@ -26,6 +26,36 @@ h2 {
     z-index: 10000;
 }
 
+/*sass-lint:disable position-relative*/
```
JPJPJPOPOP: @aditya-369 You should be using `/* sass-lint:disable class-name-format*/`. When you are using `sass-lint:disable`, you are disabling a rule, not a class/id. Also, you only need to use `/* sass-lint:disable class-name-format*/` once since you are already disabling it for the entire file.

aditya-369:
#304

PR #317, opened 2018-06-15
Fixed extraneous instant translation

### LINKED ISSUE #267: Prevent extraneous instant translation request
With instant translation enabled, if I type a word in and click translate, I see the loading spinner and it translates.

However, within ~ 2 seconds, the instant translation will trigger since I stopped typing and it will re-translate the _same_ text. This shouldn't happen since it's an extra request and a weird interface quirk.

JPJPJPOPOP:
Fixes #267

### DIFF
--- a/assets/js/translator.js
+++ b/assets/js/translator.js
@@ -6,7 +6,7 @@ var curSrcLang /*: string */, curDstLang/*: string */;
 var recentSrcLangs /*: string[] */ = [], recentDstLangs /*: string[] */ = [];
 var droppedFile/*: ?File */;
 var translateRequest;
-var translationTimer /*: TimeoutID */;
+var translationTimer/*: TimeoutID */;
 
 var UPLOAD_FILE_SIZE_LIMIT = 32E6,
     TRANSLATION_LIST_BUTTONS = 3,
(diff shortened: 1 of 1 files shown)


assets/js/translator.js
```diff
@@ -335,6 +336,8 @@ if(modeEnabled('translation')) {
                 clearTimeout(timer);
             }
 
+            translatedAlready = false;
```
sushain97: Is it possible to use `translateRequest` instead?

JPJPJPOPOP: @sushain97 Doesn't `translateRequest` go to null after translating?

sushain97: Ah, good point. I forgot what the original ticket was for. How about clearing the timeout once a translation request completes?

assets/js/translator.js
```diff
@@ -6,6 +6,7 @@ var curSrcLang /*: string */, curDstLang/*: string */;
 var recentSrcLangs /*: string[] */ = [], recentDstLangs /*: string[] */ = [];
 var droppedFile/*: ?File */;
 var translateRequest;
+var timer;
```
sushain97: Could you call it something a bit more specific like `translationTimer` since it's a global? Also, I think it can be typed.

sushain97:
This is great, thanks!

PR #319, opened 2018-06-15
Removed soft hyphens before translating/analyzing

### LINKED ISSUE #315: Remove soft hyphens before translating
There is no point in keeping them. They mess up analyses. Compare with soft hyphens:

https://www.apertium.org/index.nno.html?dir=nob-nno&q=Det%20var%20et%20problem%20i%20arbeids%C2%AD%C2%ADgruppa.#translation

and without:

https://www.apertium.org/index.nno.html?dir=nob-nno&q=Det%20var%20et%20problem%20i%20arbeidsgruppa.#translation

JPJPJPOPOP:
Fixes #315

### DIFF
--- a/assets/js/translator.js
+++ b/assets/js/translator.js
@@ -26,7 +26,7 @@ var PUNCTUATION_KEY_CODES = [46, 33, 58, 63, 47, 45, 190, 171, 49]; // eslint-di
 
 /* global config, modeEnabled, synchronizeTextareaHeights, persistChoices, getLangByCode, sendEvent, onlyUnique, restoreChoices
     getDynamicLocalization, locale, ajaxSend, ajaxComplete, localizeInterface, filterLangList, cache, readCache, iso639Codes,
-    callApy, apyRequestTimeout, isURL */
+    callApy, apyRequestTimeout, isURL, removeSoftHyphens */
 /* global ENTER_KEY_CODE, HTTP_BAD_REQUEST_CODE, HTTP_OK_CODE, SPACE_KEY_CODE, XHR_DONE, XHR_LOADING */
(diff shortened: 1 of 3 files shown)


assets/js/translator.js
```diff
@@ -723,7 +723,7 @@ function translateText(ignoreIfEmpty) {
         }
     }
 
-    var originalText /*: string */ = $('#originalText').val();
+    var originalText /*: string */ = $('#originalText').val().replace(/\u00AD/g, '');
```
sushain97: Mind adding a comment as to what's going on?

sushain97:
Huh, didn't realize there were two instances of this required. Maybe create a function in `utils.js:removeSoftHyphens` and call that? No need for a comment then.

unhammer:
Well, I have wasted way more hours than I'd like to count working around soft hyphens, and the main two issues I remember with blindly removing them were:

1. maybe some people want to keep them for pretty display in narrow columns – but I think we can ignore that (and assume anyone with fine typographical sensibilities will run their own soft-hyphenator on the output)
2. if we do any sort of matching on input/output, e.g. if we want to mark the word at index 10, then removing a soft hyphen at index 5 will turn that index to 9 – I'm pretty sure we do nothing of the sort in apertium.org

So I'm fairly sure it should be safe, and solves more bugs than it creates.

-----

One question: Is there any chance of getting an *entity* as input? Ie. `&shy;`

JPJPJPOPOP:
@unhammer I think the replace function still takes care of it. I've tried inserting &shy; in sentences and the function still removes it.

unhammer:
let's try it :)

Issue #273, opened 2018-01-30
Thumbs up and thumbs down icons for people to rate translations


ftyers:
It would be cool to have thumbs up and thumbs down icons so that people can rate translations as good/bad.

LuccoJ:
With that available, it would be useful to me to have an APy call to retrieve the average user assessment for a language pair (in order to prioritize Apertium translations over other engines I'm using in my software, depending on perceived quality).

This seems like it could be part of the work being done on #55

sushain97:
This should just be the interface through which the suggestion interaction
is started, i.e. thumbs down lets you start suggesting. I think Google does
it this way?

LuccoJ:
@sushain97 That sounds like a potentially good idea, although I'm not entirely sure... I can picture myself giving a translation a thumbs-up for being remarkable for a machine translation (maybe of a difficult sentence), but still thinking it needs refinement.

From my point of view anyway, it wouldn't "just" be that, because I could use the 0...1 number for the average thumbs in a language pair.

PR #340, opened 2019-03-03
thumbs up and thumbs down buttons for feedback are added.

sharad-99:
Solves #273


sushain97:
Thumbs up/down buttons should be added to the existing `index.html.in` and integrated via the existing JS. An entirely new HTML file is not the goal here.

sharad-99:
I'm really sorry for the mess. I am on it now.

sushain97:
No problem! Thanks for the PR.

PR #341, opened 2019-03-06
Added like and dislike buttons.

sharad-99:
solves #340


sushain97:
Please do not submit the compiled version of the HTML. The changes should be to the template HTML instead and the associated JS.

PR #353, opened 2019-12-10
Fixed dropdowns in RTL mode

### LINKED ISSUE #333: Language dropdowns bug in RTL locales
The language dropdowns in translate mode (both target and input) don't span into multiple columns on smaller screens when using RTL locale. This seems to be a JS bug.

## Expected behaviour:

Span correctly into multiple columns like on eg. English locale

apertium bug ltr

## Actual behaviour:

All languages go into one or two columns which are often longer than the viewport height

apertium bug rtl

diogoscf:
Fixes #333 

What was happening was that the available width was being calculated assuming that the source buttons were on the left and the target buttons were on the right, which is only the case if the language is ltr, so the code thought that there was much less space than there really was.

### DIFF
--- a/assets/js/translator.js
+++ b/assets/js/translator.js
@@ -27,7 +27,7 @@ var PUNCTUATION_KEY_CODES = [46, 33, 58, 63, 47, 45, 190, 171, 49]; // eslint-di
 
 /* global config, modeEnabled, synchronizeTextareaHeights, persistChoices, getLangByCode, sendEvent, onlyUnique, restoreChoices
     getDynamicLocalization, locale, ajaxSend, ajaxComplete, localizeInterface, filterLangList, cache, readCache, iso639Codes,
-    callApy, apyRequestTimeout, isURL, removeSoftHyphens, parentLang, isVariant */
+    callApy, apyRequestTimeout, isURL, removeSoftHyphens, parentLang, isVariant, langDirection */
 /* global ENTER_KEY_CODE, HTTP_BAD_REQUEST_CODE, HTTP_OK_CODE, SPACE_KEY_CODE, XHR_DONE, XHR_LOADING */
 
 if(modeEnabled('translation')) {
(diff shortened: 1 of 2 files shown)


diogoscf:
Umm... I don't understand what circleci is complaining about now...

assets/js/translator.js
```diff
@@ -554,10 +554,19 @@ function populateTranslationList() {
 
     var minColumnWidth = TRANSLATION_LIST_MAX_WIDTH / TRANSLATION_LIST_MAX_COLUMNS;
 
+    var maxSrcLangsWidth, maxDstLangsWidth;
+
     // figure out how much space is actually available for the columns
-    var maxSrcLangsWidth = $(window).width() - $('#srcLanguagesDropdownTrigger').offset().left - TRANSLATION_LISTS_BUFFER,
+    if(langDirection(locale) === 'ltr') {
```
sushain97: This issue is that Flow knows that `locale` might not be initialized by the time this function gets called.

diogoscf: Makes sense. Though how would I fix that?

sushain97: Does this code only get called after the `locale` gets initialized? If so, we should just throw here if it's `undefined` and Flow will use that as signal for its analysis.

sushain97:
> What was happening was that the available width was being calculated assuming that the source buttons were on the left and the target buttons were on the right, which is only the case if the language is ltr, so the code thought that there was much less space than there really was.

Ahhhh, right. Good catch. Probably my fault.

sushain97:
BTW, I suggest `yarn install`ing locally and running Flow/ESLint through your IDE.

assets/js/translator.js
diogoscf: I just tested and it gets called once before `locale` gets initialized and once after. But I I have thought of a way to fix this

diogoscf: Okay done

jonorthwash:
> Code LGTM. @jonorthwash, I'm guessing you've tested this?

Yep, it looks really nice (i.e., the way it's supposed to) :D

sushain97:
sorry for the extra commits, trying to get dockerhub to build commits again and it seems to have worked!

Issue #390, opened 2021-02-01
new features meta issue


jonorthwash:
The following 
* [ ] dictionary lookup (#105)
* [ ] spell checking (#253)
* [ ] suggestions (#55, #252)
* [ ] multi-step translation (#91, #120)

tahsinzisan:
> The following features constitute major improvements to apertium-html-tools that have been worked on but are not finished:
> 
> * [ ]  dictionary lookup ([Dictionary lookup #105](https://github.com/apertium/apertium-html-tools/issues/105))
> * [ ]  spell checking ([Spell Checker Interface [WIP] #253](https://github.com/apertium/apertium-html-tools/pull/253))
> * [ ]  suggestions ([Suggestions interface #55](https://github.com/apertium/apertium-html-tools/issues/55), [Issue 55 suggestions 2 #252](https://github.com/apertium/apertium-html-tools/pull/252))
> * [ ]  multi-step translation ([Chained translation path display #91](https://github.com/apertium/apertium-html-tools/issues/91), [mark destinations through multi-step translation somehow #120](https://github.com/apertium/apertium-html-tools/issues/120))

Sir, I got to know about Apertium recently and I am willing to contribute to open source project under apertium. I am a beginner and don't have any advanced skills on anything but I know html, css, javascript, python, java, c++, c to some to basic and some to intermediate. I am learning git and github thoroughly. I'll be more than happy to learn new frameworks. I really want to see my knowledge gets implemented in real life. So please guide me how and where should I start and suggest me any beginner friendly project that matches with my skills. 
I sincerely apologize for any unintentional inappropriate behavior
My email: tahsinzisan5@gmail.com

tahsinzisan:
> The following features constitute major improvements to apertium-html-tools that have been worked on but are not finished:
> 
> * [ ]  dictionary lookup ([Dictionary lookup #105](https://github.com/apertium/apertium-html-tools/issues/105))
> * [ ]  spell checking ([Spell Checker Interface [WIP] #253](https://github.com/apertium/apertium-html-tools/pull/253))
> * [ ]  suggestions ([Suggestions interface #55](https://github.com/apertium/apertium-html-tools/issues/55), [Issue 55 suggestions 2 #252](https://github.com/apertium/apertium-html-tools/pull/252))
[...]

PR #252, opened 2018-01-04
Issue 55 suggestions 2

sushain97:

Issue #55, opened 2016-05-16
Suggestions interface


svineet:
The suggestions work on the backend was merged into the suggestions branch in this PR: https://github.com/goavki/apertium-apy/pull/18

Overall, the suggestions route works and is almost production ready, but the following needs to be taken care of (feel free to add anything in the comments):
* The UI in html tools needs refinement. spectie and firespeaker had some ideas, we could discuss this here.
* Tests could be written for apy endpoints.
* The whole mechanism could use a lot of rigorous testing, there might be bugs.
* ReCAPTCHA API supports localizations and I was thinking of integrating that during the GCI task, but it slipped out of my mind, could be a great idea.

Current interface is at https://github.com/svineet/apertium-html-tools/tree/suggest-word and some of that work was merged into https://github.com/goavki/apertium-html-tools/tree/suggestions but my later improvements still have open PRs to the suggestions branch of html-tools. So we need to work on organizing that stuff.

jonorthwash:
I merged the two open suggestions PRs (https://github.com/goavki/apertium-html-tools/pull/35, https://github.com/goavki/apertium-html-tools/pull/37) into the suggestions branch.  Now @Kira-D should go through your list of to-do points.

Kira-D:
@svineet can you, please, give me more information about the UI refinement?
What sort of tests do you have in mind?

sushain97:
moved from PR:

@shardulc @jonorthwash I went ahead and set it up on my server (with a recaptcha key) so we can easily test: https://azurite.skc.name/apertium-html-tools/build/index.eng.html

You need to check `Mark unknown words` for it to take effect (type some nonsense to translate and you'll see the red).

There are definitely some UI touches that need to take place but I haven't thoroughly tested it yet. Whoever can get to it first should take a crack at it!

I also have a hunch that it's making my CPU spike... Here be dragons. Would appreciate it if someone could confirm that it's not just me.

shardulc:
@sushain97 When I click the suggest button, nothing happens and I cannot see any network activity either. (The captcha works fine.) However, I can't debug this because I only see `min.js` :smiley:

sushain97:
Let me run make debug next time I'm not afk and it'll be easier.

sushain97:
@shardulc https://azurite.skc.name/apertium-html-tools/build/index.debug.html

Don't try too much to debug it ;) Just make a list of problems.

shardulc:
List of problems:
[...]

PR #395, opened 2021-03-24
Issue 55 suggestions 3

thehanimo:
Rebases changes from #252.
Fixes reCAPTCHA re-render on locale change.
Fixes suggestions modal tooltip errors

Related Issues: #55, #390


sushain97:
The build failure indicates that you need to run `make` in `assets/strings`.

thehanimo:
Thanks @sushain97! Btw, I had to change this:

https://github.com/apertium/apertium-html-tools/blob/e6226cc6dc475590abaa92ada1e778ae1f5f7a34/assets/strings/Makefile#L3

to:

```sh
cleanup: $(shell find . -regex ".*/[a-z]\{3\}\.json")
```

Don't think -regextype and -printf is supported on Darwin Kernel Version 20.3.0

TinoDidriksen:
That would be BSD find vs. GNU find. A portable solution is something like `find . -name '*.json' | egrep '/[a-z]{3}\.json'` but should be tested.

`localizeReacaptchaAlternativeLanguages` has a typo in both places it is used.

But asides from that the code looks ok to me.

thehanimo:
Nice catch!

assets/js/localization.js
```diff
@@ -1,27 +1,35 @@
...
+var localizeRecaptchaLanguages = ['ar', 'af', 'am', 'hy', 'az', 'eu', 'bn', 'bg', 'ca', 'zh-HK', 'zh-CN', 'zh-TW', 'hr', 'cs', 'da', 'nl', 'en-GB', 'en', 'et', 'fil', 'fi', 'fr', 'fr-CA', 'gl', 'ka', 'de', 'de-AT', 'de-CH', 'el', 'gu', 'iw', 'hi', 'hu', 'is', 'id', 'it', 'ja', 'kn', 'ko', 'lo', 'lv', 'lt', 'ms', 'ml', 'mr', 'mn', 'no', 'fa', 'pl', 'pt', 'pt-BR', 'pt-PT', 'ro', 'ru', 'sr', 'si', 'sk', 'sl', 'es', 'es-419', 'sw', 'sv', 'ta', 'te', 'th', 'tr', 'uk', 'ur', 'vi', 'zu'];
+var localizeRecaptchaAlternativeLanguages = {'zh': 'zh-TW', 'hrv': 'hr', 'srp': 'sr', 'msa': 'ms', 'cy': 'en', 'ga': 'en', 'gv': 'en', 'gd': 'en', 'rn': 'en', 'lg': 'en', 'mt': 'en', 'sq': 'fr', 'br': 'fr', 'se': 'no', 'eo': 'en', 'fo': 'da', 'ia': 'en', 'xh': 'af', 'mfe': 'fr', 'nb': 'no', 'nn': 'no', 'uz': 'ru', 'rm': 'en', 'tg': 'ru', 'ba': 'ru', 'be': 'ru', 'os': 'ru', 'kum': 'ru', 'ky': 'ru', 'mk': 'en', 'tt': 'ru', 'kk': 'ru', 'he': 'en', 'ne': 'hi', 'as': 'hi', 'pa': 'ur', 'mrj': 'ru', 'myv': 'ru', 'oc': 'es', 'cv': 'ru', 'arg': 'es', 'ast': 'es', 'hbs': 'en', 'bos': 'en', 'nog': 'ru', 'sah': 'ru', 'uig': 'zh-TW', 'tyv': 'ru'};
```
jonorthwash, line 8: Does this list need to be updated manually?  If so, where from, and how often?

assets/strings/eng.json
```diff
@@ -54,6 +54,10 @@
     "Download": "Download",
     "Contact": "Contact",
     "Documentation": "Documentation",
+    "Suggest_Title": "Improve Apertium's translation",
+    "Suggest_Button": "Suggest",
+    "Suggest_Sentence": "How would you suggest we translate {{targetWordCode}}?",
+    "Suggest_Placeholder": "New Word",
```
jonorthwash, line 60: It won't always be a new word; in fact, often the problem could be something more having to do with grammar.

jonorthwash (changes requested):

assets/strings/eng.json
thehanimo: How about "Suggested word" or "Your suggestion" instead?

assets/js/localization.js
thehanimo: `localizeRecaptchaLanguages` is the list of languages reCAPTCHA [supports.](https://developers.google.com/recaptcha/docs/language)
`localizeRecaptchaAlternativeLanguages` provides a fallback language in case the locale (current website language) is not supported by reCAPTCHA.

We would need to update either/both when we support new locales or when reCAPTCHA supports more/less languages

jonorthwash: Could you put a comment to that effect here?

assets/strings/eng.json
thehanimo: #252 had a Suggest_Placeholder value in Russian as well (Новое слово). Changed that to "Ваше предложение" based on Google Translate and IRC chat.

thehanimo:
To highlight individual words and to make them clickable,
```html
<textarea class="form-control bg-light" id="translatedText" rows="15" spellcheck="false"
aria-label="Translated Text" readonly></textarea>
```
was replaced with 
```html
<div class="form-control bg-light" id="translatedText" rows="15" spellcheck="false"
aria-label="Translated Text" readonly></div>
```
in #252

The only disadvantage I see with this is the missing resize-handle in the translated text container, which can be fixed by setting style to `resize: vertical; overflow: auto;`. But hovering over the handle still shows a text cursor (set for the div), rather than a resize one. Maybe we can add another transparent div at the bottom corner with a `cursor: se-resize` to fix that.

Apart from that, I'm not sure if this has far-reaching effects.

Issue #376, opened 2020-08-16
Swap button should swap text as well


khannatanmai:
The swap button on the website should swap the text in the text boxes as well so that the button can be used for back translation as it is on Google Translate. Right now there's not much use of the button.

jonorthwash:
I think an argument could be made for either behaviour.  I have need of both from time to time, so am happy either way.

khannatanmai:
You would need to just switch the languages and not the text if you somehow by mistake put text in the wrong box or something. The argument for switching the text as well, apart from back translation could also be if I wanna translate from A to B to C, so I translate A to B, then switch and then change target language to C. Is there some use case I'm missing for the first functionality?

sushain97:
I'm +1 to us behaving the same way that Google Translate does. I imagine they've done some amount of user research to figure out what behavior folks find most useful/intuitive.

CipherGirl:
I would like to try working on this issue. I am a GSoC aspirant.

PR #396, opened 2021-03-29
Swap text with swap button

CipherGirl:
The swap button will now swap the texts of the textarea boxes.

Fixes: #376

### DIFF
--- a/assets/js/translator.js
+++ b/assets/js/translator.js
@@ -220,6 +220,10 @@ if(modeEnabled('translation')) {
                 refreshLangList(true);
                 muteLanguages();
 
+                var temp = $('#originalText').val();
+                $('#originalText').val($('#translatedText').val());
+                $('#translatedText').val(temp);
+
                 if($('.active > #detectedText')) {
                     $('.srcLang').removeClass('active');
                     $('#srcLang' + (recentSrcLangs.indexOf(curSrcLang) + 1)).addClass('active');


jonorthwash (changes requested):
It's perhaps poor memory management to not empty or destroy `temp`.  Otherwise the code looks okay.  I'm also curious about @sushain97's thoughts, though.

TinoDidriksen:
In 99% of cases, you don't need to even think about memory management in JS. This is definitely such a case. Garbage collection will take care of that one.

CipherGirl:
How to resolve the build test?

TinoDidriksen:
Run `eslint --config assets/js/.eslintrc.json assets/js/*.js` on your code and fix everything it complains about. In this case, it's simply that you have lines with space before the terminator.

TinoDidriksen (approved):

sushain97 (approved):

unhammer:
stars

Looks like some race condition, like the text is sent through the old language pair and thus gets stars. And the stars keep getting added …

sushain97:
Thanks for reporting! Reverted this PR for now.

TinoDidriksen:
It is live on https://beta.apertium.org/, and I cannot reproduce that race condition.

TinoDidriksen:
And there's no way it should be able to race, because setting the value via JS doesn't trigger the on-change event.

TinoDidriksen:
And nothing in the code https://github.com/apertium/apertium-html-tools/blob/master/assets/js/translator.js#L201-L232 triggers the translation again either. On Chrome and Firefox, F12 tools show no requests regardless of how many times the Swap button is used. Cannot reproduce.

TinoDidriksen:
Undid the undo.

PR #401, opened 2021-04-09
Change ESLint comma-dangle to always-multiline, except for function args

TinoDidriksen:
The diff from https://github.com/apertium/apertium-html-tools/pull/400 has superfluous line changes, which made me realize this project enforces no trailing commas. So change that to enforce trailing commas, except for function args, as this reduces diff overhead and copy-paste errors. It is a travesty that JSON doesn't allow trailing commas, but we should absolutely require it for all JS lists.

### DIFF
--- a/assets/js/.eslintrc.json
+++ b/assets/js/.eslintrc.json
@@ -9,7 +9,15 @@
 		"array-callback-return": "error",
 		"block-spacing": "error",
 		"camelcase": "error",
-		"comma-dangle": "error",
+		"comma-dangle": [
+			"error", {
+				"arrays": "always-multiline",
+				"objects": "always-multiline",
+				"imports": "always-multiline",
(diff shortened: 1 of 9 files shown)


TinoDidriksen:
We should also set https://eslint.org/docs/rules/no-var in the future, though that requires a lot more testing to ensure all scopes are actually correct.

sushain97 (approved):
Historically, it's been this way because we supported old versions of IE. Since we only support whatever Bootstrap supports, we now support IE 10 which I think accepts trailing commas: https://caniuse.com/?search=trailing.

I can't really check though so I suggest double checking before merging.

TinoDidriksen:
That makes sense. IE11 is the only one anyone needs to care about these days, but IE11 is unfortunately still relevant. Luckily, IE11 supports trailing commas and `let`.

Issue #414, opened 2021-06-15
Add noscript warning


flammie:
I use noscript extension on chromium and beta.apertium.org opens currently totally empty screen. At least a short informative message would be nice, but I think that just translating plain text snippet should not really need javascript...

sushain97:
Happy to approve a PR which adds a noscript tag.

It has never been an objective of this repository to support no-JS functionality and I don't expect that it ever will be given its feature set and goals.

flammie:
sounds good, I'm not very good at modern web stuffs but if I read the code correctly it's probably not easy to make the noscript part localisable?

sushain97:
> it's probably not easy to make the noscript part localisable?

I can't see how that would be possible since you can't change the text or redirect the user based on their locale without using JS.

PR #416, opened 2021-06-17
noscript tag

flammie:
fix for issue #414

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -21,4 +21,5 @@
 <body>
   <div id="react-mount"></div>
   <script type="text/javascript" src="./index.js?{{CACHE_BREAK}}"></script>
+  <noscript>You cannot use apertium without javascript!</noscript>
 </body>
--- a/src/index.html
+++ b/src/index.html
@@ -21,5 +21,5 @@
 <body>
   <div id="react-mount"></div>
   <script type="text/javascript" src="./index.js?{{CACHE_BREAK}}"></script>
-  <noscript>You cannot use apertium without javascript!</noscript>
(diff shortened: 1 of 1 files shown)


src/index.html
```diff
@@ -21,4 +21,5 @@
 <body>
   <div id="react-mount"></div>
   <script type="text/javascript" src="./index.js?{{CACHE_BREAK}}"></script>
+  <noscript>You cannot use apertium without javascript!</noscript>
```
sushain97, line 24:
```suggestion
  <noscript>You cannot use this website without JavaScript!</noscript>
```

sushain97 (approved):

sushain97 (approved):

unhammer:
Are there non-js sites we could link to? ( http://aplica.prompsit.com/en/ doesn't seem to work any more. ) or maybe just a link to https://wiki.apertium.org/wiki/Interfaces and non-js users might choose to install packages or simpleton etc.

xavivars:
Does anyone have a sense of the % of people that may be browsing without JS?

Few years ago, I would have pushed not only to have a link, but even to provide a non-JS alternative. But with all the improvements that JS has brought, the JS-framework adoption & SPAs approach that most sites have taken, it's hard to "browse internet" with JS disabled.

unhammer:
Probably not many, but [once in a while it can be essential to someone](https://shkspr.mobi/blog/2021/01/the-unreasonable-effectiveness-of-simple-html/). If anyone actually makes a non-js version, we should at least link to it in the noscript tag :)

PR #418, opened 2021-07-09
Target ES2019 (fixes #417)

### LINKED ISSUE #417: website broken with latest palemoon browser
website is completely blank, js console reveals

```
SyntaxError: expected expression, got '?'[Learn More]
in index.js: return(e=e?e.displayName||e.name:"")
```

some web developer to eager to use latest JS features?

TinoDidriksen:
As per https://esbuild.github.io/api/#target default is `esnext` which is way too new. Target should be $YEAR - 5, currently ES2016.

### DIFF
--- a/build.ts
+++ b/build.ts
@@ -189,6 +189,7 @@ void esbuild.build({
   entryPoints: ['src/index.tsx'],
   bundle: true,
   metafile: true,
+  target: 'es2019',
   loader: { '.embed.png': 'dataurl', '.png': 'file', '.gif': 'file', '.jpg': 'file' },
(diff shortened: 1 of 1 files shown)


sushain97:
> Target should be $YEAR - 5, currently ES2016.

Are there any browsers versions from 2016 which are not EOL? Given how fast browser vendors release security patches, I'd be pretty surprised. Depending on impact to bundle size, I'd prefer starting somewhere like ES2019 instead.

TinoDidriksen:
Sure, sounds good. Looks like distros package recent browsers even for stable old distros.

Issue #361, opened 2020-01-21
add string directing to beta.apertium.org


jonorthwash:
The main www.apertium.org website should have a string somewhere (in the "about" modal?  as an MOTD?) to the effect of "Looking for more languages?  Try [beta.apertium.org](http://beta.apertium.org)."

The implementation should be general enough to allow other sites that use html-tools to display it or not, and also for it to be localised.

jonorthwash:
I believe this is now addressed.

PR #422, opened 2021-10-02
Add string replacements and the string in the about modal

Devesh21700Kumar:
The issue #361 

1).Added the string and created a reference instance "More_Languages" in the eng.json file along with a replacement string "more_languages" to use the anchor tag.

2). Included it  the tests as well

image


sushain97:
This seems like it would result in both `beta.apertium.org` and `apertium.org` having this string. That seems like a bit of a confusing UX?

Devesh21700Kumar:
> This seems like it would result in both `beta.apertium.org` and `apertium.org` having this string. That seems like a bit of a confusing UX?

I don't think that `apertium.org` would have the same string as  beta.apertium.org has a separate string instance named `More_Languages` in the eng.json file. Also, even the text wrapped around the anchor tag is `beta.apertium.org` so I guess that won't be an issue at all.

sushain97:
Ah, my bad. I misread the implementation.

Devesh21700Kumar:
> Ah, my bad. I misread the implementation. In this case, how would `beta.apertium.org` avoid having this string entirely? It uses the same source tree as `apertium.org`.

Yeah, this was something that I thought about as well. But then the main reason for using the string is to encapsulate the whole `Looking for more languages? Try {{more_languages}}` part in `{{More_Languages}}`.


Here
`{{more_languages}}` is `<a dangerouslySetInnerHTML={{ __html: 'beta.apertium.org' }} href='https://beta.apertium.org' rel='noreferrer'  target='_blank'/>`

and since this appeared quite modular I thought of going ahead with this implementation.

Devesh21700Kumar:
> > Ah, my bad. I misread the implementation. In this case, how would `beta.apertium.org` avoid having this string entirely? It uses the same source tree as `apertium.org`.
> 
> Yeah, this was something that I thought about as well. But then the main reason for using the string is to encapsulate the whole `Looking for more languages? Try {{more_languages}}` part in `{{More_Languages}}`.
> 
> Here `{{more_languages}}` is `<a dangerouslySetInnerHTML={{ __html: 'beta.apertium.org' }} href='https://beta.apertium.org' rel='noreferrer' target='_blank'/>`
> 
> and since this appeared quite modular I thought of going ahead with this implementation.

@sushain97  any follow-up on this or any workaround, you would like to suggest?

sushain97:
Maybe you could add a configuration option that enabled or disabled showing this string?

@jonorthwash requested this feature so perhaps he has an opinion on whether it matters that we show this string on both `apertium.org` and `beta.apertium.org`?

Devesh21700Kumar (commented):
@sushain97 sorry for the very late follow up.
To end all confusions I have added a logic to check the path of the route and conditionally render the string.

sushain97:
> To end all confusions I have added a logic to check the path of the route and conditionally render the string.

Rather than casing on the location within the frontend, we should have a configuration option that conditionally renders the URL.

PR #440, opened 2022-03-11
Support pairprefs in doc and web translation

unhammer:
https://github.com/apertium/apertium-html-tools/issues/439


sushain97 (approved):
Added a couple extra tests. Thanks for doing the hard part!

PR #457, opened 2023-02-19
Tweak style prefs layout

sushain97:
Hopefully fixes https://github.com/apertium/apertium-html-tools/issues/435 (sorry for the delay on this).

image

image

It isn't the _greatest_ but I think it's reasonable enough?

### DIFF
--- a/src/components/translator/TranslationOptions.tsx
+++ b/src/components/translator/TranslationOptions.tsx
@@ -69,8 +69,8 @@ const TranslationOptions = ({
           <Col
             style={{
               maxHeight: '12rem',
-              overflowY: 'scroll',
               maxWidth: 'calc(min(100vw - 30px, 400px))',
+              overflowY: 'scroll',
               width: 'max-content',
             }}
           >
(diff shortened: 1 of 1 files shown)


unhammer:
Yeah, that helps :) The scroll bar indicator gets almost invisible when I make it mobile-sized, but maybe it's more visible on an actual phone? Not a big thing though, this is a big enough improvement that I'd like to get it merged asap :)

sushain97:
> I see it has a max-height, which I can increase, is it set kind of low just in case it goes outside the viewport or whatever it's called?

Yeah, the idea here was just to keep it roughly inside the viewport for standard screen sizes.

> The scroll bar indicator gets almost invisible when I make it mobile-sized, but maybe it's more visible on an actual phone?

I gave it a look and it seems alright! Good flag.

> Not a big thing though, this is a big enough improvement that I'd like to get it merged asap :)

👍 A larger change we could consider is having it open in a modal. I'm not sure if that would be annoying or not based on how this element gets used in practice. Another variant of the modal approach is an [Offcanvas](https://getbootstrap.com/docs/5.0/components/offcanvas/#live-demo) element.

unhammer:
In practice I'm guessing people would set the stuff once (with perhaps a tiny bit of later tweaking) and once set they probably won't touch it much again. But when setting it they need to hunt-and-peck through 100 entries. So a bit like typical Preferences menus :) Offcanvas would probably be nice here, taking up more space on the screen (and on desktop has the added benefit over modal of letting you immediately see the changes in the output, though that's not a must-have)

PR #460, opened 2023-03-03
Update locales.json

mpQadir:
Saraiki added


sushain97:
This is already complete: https://github.com/apertium/apertium-html-tools/blob/master/src/strings/locales.json#L27.

Issue #462, opened 2023-03-09
Google Summer of code in the about section behaves as a link


456535688765:
What Happens: The Google summer of code acts like a link instead of staying unchanged.
while hovering it changes its state and it gets underlined

Solution : need some changes in the img section

TinoDidriksen:
The actual bug is that https://summerofcode.withgoogle.com/static/img/summer-of-code-logo.svg is now a 404.

Could replace it with https://summerofcode.withgoogle.com/assets/media/logo.svg though depending on 3rd party images is a bad idea.

456535688765:
should I replace it with https://summerofcode.withgoogle.com/assets/media/logo.svg this image

TinoDidriksen:
Looks like other assets are in https://github.com/apertium/apertium-html-tools/tree/master/src/components/footer/img, so I'd say it should go there.

456535688765:
So do I make a change in the code base or need to upload the image file here  https://github.com/apertium/apertium-html-tools/tree/master/src/components/footer/img

456535688765:
Ok Got it Adding the Google summer of code img to this section https://github.com/apertium/apertium-html-tools/tree/master/src/components/footer/img

456535688765:
Sharing the PR 
Thank you @TinoDidriksen for your help

456535688765:
Hey @TinoDidriksen can you help me with the merging process

hazemk537:
help in merge plz,, https://github.com/apertium/apertium-html-tools/pull/464

456535688765:
Thhe issue is not solved yet @hazemk537
![Screenshot (230)](https://user-images.githubusercontent.com/92295764/224478140-774a8a55-966f-4451-a9f8-316a3b56ecb7.png)

hazemk537:
@456535688765 we can do it inshaallah

456535688765:
I have resolved this Google Summer of code in the about section behaves as a link #462 problem but getting issues with the formatting of the code

456535688765:
Can you help me @TinoDidriksen?

456535688765:
![Screenshot (232)](https://user-images.githubusercontent.com/92295764/224491788-ba8fed67-5bd8-4735-a786-3b6140cd7f6d.png)
resolved

456535688765:
should I start with the PR

TinoDidriksen:
Yes, I don't know why you closed the PR.

[...]

PR #465, opened 2023-03-11
Google Summer of code in the about section behaves as a link #462

456535688765:
resolved The bug where the image was not visible


hazemk537:
why this pull request is waiting,sir?

TinoDidriksen:
It doesn't solve the problem in the correct way, in my opinion. The asset should be copied into the repo to avoid any future 404, same as the other images are. And setting both width and height shouldn't be necessary, as constraining just one dimension will let the browser scale it automatically.

PR #467, opened 2023-03-12
issue#462 about section

hazemk537:


sushain97 (changes requested):
This looks reasonable but it's not clear to me why you've added both a PNG and an SVG version. The SVG is preferable since it's significantly smaller.

hazemk537:
Because , I found an error while building when I used the SVG version , so I convert it to PNG

sushain97:
If you run into an unexpected issue, please ask for help with it rather than reverting to a sub-optimal choice. In this case, you'll need to modify Jest config, ESBuild config and TS declarations to allow for SVG files.

hazemk537:
okay , I can do that

hazemk537:
should i install esbuild-plugin-svg and use it ?
it is deprecated

sushain97:
Please do not add any new config files or new dependencies. The existing support for gif/png files should extend to svg files immediately.

hazemk537:
# After many tries 
- I cannot see the svg file in the dist folder .
- I searched and found that 
1. The esbuild built-in loader support s GIF  , PNG  and JPEG and donot support SVG.
2.  but  If you want to load SVG files ,You should use extension like esbuild-plugin-svg

# References
[If you want to load PNG, JPEG or GIF files as data URLs in your JavaScript code, you can use esbuild’s built-in loader option](https://blog.logrocket.com/getting-started-esbuild/)

[If you want to load SVG files as strings or custom elements in your JavaScript code, you can use esbuild-plugin-svg](https://github.com/nativew/esbuild-plugin-svg)

TinoDidriksen:
We do NOT want any of the images as data URLs. But surely it can't be true that we need something special to copy an asset from the source to the build folder and insert the relative path to it.

hazemk537:
okay , should I repo the repo to PNG version ? to accept the pull request ?

TinoDidriksen:
No, the SVG version is correct. It should simply be handled same as the other images are - which is, copied to the build folder and linked from there.

hazemk537:
No need to copy to dist , its working now using the SVG !!
image

src/components/footer/img/GSOC_logo.png
sushain97, line 1: Can this file be deleted now?

src/globals.d.ts
```diff
@@ -12,3 +12,8 @@ declare module '*.jpg' {
   const url: string;
   export default url;
 }
+
+declare module '*.svg' {
+  const url: string
+  export default url;
+}
```
sushain97, line 19: Please add back the trailing newline.

sushain97 (changes requested):

hazemk537:
Done

sushain97:
Please fix the build failures.

hazemk537:
done

sushain97:
I don't think we should be increasing the test timeout. It's probably just flaky? Another run will fix it.

hazemk537:
I tried runnibg it using jest and without increasing the timeout ,one file sometimes fails but after the increasing , no fails .
I think we shoud increase it for future non fake  issues during build .
do you agrre with me ?

sushain97:
No, we should fix the test to not flake instead.

hazemk537:
so the problem is inside the test file itself ! I got it

sushain97 (approved):

PR #481, opened 2023-12-08
fixes #477 (Build should not rely on git)

### LINKED ISSUE #477: Build should not rely on git
`git describe --tags --always` at https://github.com/apertium/apertium-html-tools/blob/master/build.ts#L65 and potentially other places should be removed. A build must not rely on vcs tools.

Also, the build instructions [don't say to install git](https://github.com/orgs/apertium/discussions/40).

satti-hari-krishna-reddy:
fixes #477 

**Description :**

This pull request addresses the issue where the build process was relying on Git for version information. The problem arose when the build process did not have Git available.

**Changes Made :**

  -In package.json, updated the `version` field to the latest released GitHub version

 -In build.ts, modified the code to use the version specified in package.json.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "author": "Sushain Cherivirala <sushain@skc.name>",
   "license": "GPL-3.0-or-later",
   "scripts": {
-    "build": "ts-node build.ts",
+    "build": "BUILD_VERSION=$BUILD_VERSION ts-node build.ts",
     "serve": "python3 -m http.server --directory dist",
     "tsc": "tsc --noEmit",
     "eslint": "eslint . --ext .ts,.tsx --max-warnings 0",
--- a/package.json
+++ b/package.json
@@ -1,10 +1,10 @@
 {
   "name": "apertium-html-tools",
-  "version": "5.0.0",
+  "version": "5.1.8",
   "author": "Sushain Cherivirala <sushain@skc.name>",
   "license": "GPL-3.0-or-later",
   "scripts": {
-    "build": "BUILD_VERSION=$BUILD_VERSION ts-node build.ts",
+    "build": "ts-node build.ts",
     "serve": "python3 -m http.server --directory dist",
     "tsc": "tsc --noEmit",
     "eslint": "eslint . --ext .ts,.tsx --max-warnings 0",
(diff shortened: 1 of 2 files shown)


satti-hari-krishna-reddy:
@sushain97 could you please review my changes? Your feedback would be highly appreciated.

package.json
```diff
@@ -4,7 +4,7 @@
   "author": "Sushain Cherivirala <sushain@skc.name>",
   "license": "GPL-3.0-or-later",
```
sushain97, line 5: Can you update the `version` field above to actually match the latest released GH version if we're going to use it as the canonical source?

sushain97 (changes requested):
Seems reasonable

satti-hari-krishna-reddy:
@sushain97 sorry for late reply college work has been busy, Sure, I'll update the version in package.json field above to actually match the latest released git hub version but i got a doubt , Since we're moving away from direct reliance on Git ,will manual updates to package.json be required for every  release?

TinoDidriksen:
> ...will manual updates to package.json be required for every release?

Yes. That is the point. It's about what the source of truth is, and I hold that it should be the code - GitHub is merely storage, and must not be an integral part of the project.

sushain97:
> Since we're moving away from direct reliance on Git ,will manual updates to package.json be required for every release?

Yes, this is fine. Ideally, we set up a GH workflow to create a tag and it'll fail when the tag already exists and succeed when a commit is pushed to `master` with a new tag.

sushain97:
> Yes. That is the point. It's about what the source of truth is, and I hold that it should be the code - GitHub is merely storage, and must not be an integral part of the project.

I don't disagree with this but I'd argue GitHub isn't relevant here. The tags are a Git construct which is arguably part of the source of truth. The point that a build shouldn't require `.git` is one I agree with though so this is fine in general.

satti-hari-krishna-reddy:
Hey @sushain97 , @TinoDidriksen , I've made some changes to the code. Can you take a look and let me know if everything's good now? Thanks!

sushain97 (dismissed):

sushain97:
@satti-hari-krishna-reddy CI is failing.

satti-hari-krishna-reddy:
@sushain97 , I've got the Prettier error fixed, so the CI should be good to go now!

sushain97 (approved):

PR #482, opened 2024-01-02
 Improve swap lang button for language and text swapping

### LINKED ISSUE #469: Switching languages doesn't copy already provided input and output
IMO, using the current output as the new query would greatly improve UX.

Instead of `https://beta.apertium.org/#?dir=spa-eng&q=Buenos%20Muchacho`
to `https://beta.apertium.org/#?dir=eng-spa&q=Buenos%20Muchacho`, 

``https://beta.apertium.org/#?dir=spa-eng&q=Buenos%20Muchacho` to `https://beta.apertium.org/#?dir=eng-spa&q=Good%20Boy`?

satti-hari-krishna-reddy:
fixes #469 

**Description :**

This pull request addresses the issue where switching languages doesn't copy already provided input and output. With the modified changes now  when  the languages are swapped the new input is our already provided output.

before : `https://beta.apertium.org/index.eng.html#?dir=eng-spa&q=Dog` to `https://beta.apertium.org/index.eng.html#?dir=spa-eng&q=Dog`

After : `https://beta.apertium.org/index.eng.html#?dir=eng-spa&q=Dog` to `https://beta.apertium.org/index.eng.html#?dir=spa-eng&q=Perro`

**Changes Made :**
 -Lifted state from the child component `TextTranslationForm` to the parent component `Translator`, passing the state and its corresponding setter functions as props and wrapped setSrcText and setTgtText in the function swapLangText to swap the text and passed it to LanguageSelector as a prop. Included that function in swapLangs function to call it whenever it is executed.

 -Modified `TextTranslationForm.test` to accommodate the new changes made to `TextTranslationForm`. In `LanguageSelector.test`, checked whether `swapLangText` is actually getting called or not.

### DIFF
--- a/src/components/translator/Translator.tsx
+++ b/src/components/translator/Translator.tsx
@@ -243,11 +243,16 @@ const WithTgtLang = ({
 const Translator = ({ mode: initialMode }: { mode?: Mode }): React.ReactElement => {
   const mode: Mode = initialMode || Mode.Text;
 
+  const textUrlParam = 'q';
   const { t } = useLocalization();
   const history = useHistory();
   const config = React.useContext(ConfigContext);
 
   const [loading, setLoading] = React.useState(false);
+  const [srcText, setSrcText] = useLocalStorage('srcText', '', {
+    overrideValue: getUrlParam(history.location.search, textUrlParam),
+  });
+  const [tgtText, setTgtText] = React.useState('');
 
   const [markUnknown, setMarkUnknown] = useLocalStorage('markUnknown', false);
   const [instantTranslation, setInstantTranslation] = useLocalStorage('instantTranslation', true);
@@ -271,6 +276,11 @@ const Translator = ({ mode: initialMode }: { mode?: Mode }): React.ReactElement 
   }
 
   const onTranslate = React.useCallback(() => window.dispatchEvent(new Event(TranslateEvent)), []);
+
+  const swapLangText = () => {
+    setSrcText(tgtText);
+    setTgtText(srcText);
+  };
 
   return (
     <Form
@@ -307,12 +317,24 @@ const Translator = ({ mode: initialMode }: { mode?: Mode }): React.ReactElement 
                     tgtLang,
                     detectedLang,
                     loading,
+                    swapLangText,
                   }}
                 />
                 {(mode === Mode.Text || !mode) && (
                   <>
                     <TextTranslationForm
-                      {...{ instantTranslation, markUnknown, setLoading, srcLang, tgtLang, pairPrefs }}
+                      {...{
+                        instantTranslation,
+                        markUnknown,
+                        setLoading,
+                        srcLang,
+                        tgtLang,
+                        pairPrefs,
+                        srcText,
+                        tgtText,
+                        setSrcText,
+                        setTgtText,
+                      }}
                     />
                     <Row className="mt-2 mb-3">
                       <Col className="d-flex d-sm-block flex-wrap translation-modes" md="6" xs="12">
--- a/src/components/translator/__tests__/LanguageSelector.test.tsx
+++ b/src/components/translator/__tests__/LanguageSelector.test.tsx
@@ -23,6 +23,7 @@ const renderLanguageSelector = (props_: Partial<Props> = {}): Props => {
     detectLangEnabled: true,
     detectedLang: null,
     setDetectedLang: jest.fn(),
+    swapLangText: jest.fn(),
     ...props_,
   };
 
@@ -90,7 +91,7 @@ describe('swapping', () => {
   });
 
   it('allow swapping when swapped pair valid', () => {
-    const { srcLang, tgtLang, setSrcLang, setTgtLang } = renderLanguageSelector({
+    const { srcLang, tgtLang, setSrcLang, setTgtLang, swapLangText } = renderLanguageSelector({
       tgtLang: 'spa',
     });
 
@@ -98,6 +99,7 @@ describe('swapping', () => {
 
     expect(setSrcLang).toHaveBeenCalledWith(tgtLang);
     expect(setTgtLang).toHaveBeenCalledWith(srcLang);
+    expect(swapLangText).toHaveBeenCalled();
   });
 });
 
--- a/src/components/translator/__tests__/TextTranslationForm.test.tsx
+++ b/src/components/translator/__tests__/TextTranslationForm.test.tsx
@@ -7,6 +7,10 @@ import userEvent from '@testing-library/user-event';
 
 import { DetectCompleteEvent, DetectEvent, TranslateEvent } from '..';
 import TextTranslationForm, { Props } from '../TextTranslationForm';
+import { getUrlParam } from '../../../util/url';
+import useLocalStorage from '../../../util/useLocalStorage';
+
+const textUrlParam = 'q';
 
 const renderTextTranslationForm = (
   props_: Partial<Props> = {},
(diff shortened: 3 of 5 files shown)


satti-hari-krishna-reddy:
@sushain97 , could you please review my pull request?

sushain97:
This looks reasonable at first glance but I'll take a closer look tomorrow. I may have been wrong about the cleanest way to do this since I haven't looked at the code in a while. I appreciate the detailed PR description.

src/components/translator/Translator.tsx
```diff
@@ -243,11 +243,16 @@ const WithTgtLang = ({
 const Translator = ({ mode: initialMode }: { mode?: Mode }): React.ReactElement => {
   const mode: Mode = initialMode || Mode.Text;
 
+  const textUrlParam = 'q';
```
sushain97, line 246: This shouldn't be repeated twice. It should be defined in exactly one place.

src/components/translator/Translator.tsx
```diff
@@ -272,6 +277,11 @@ const Translator = ({ mode: initialMode }: { mode?: Mode }): React.ReactElement
 
   const onTranslate = React.useCallback(() => window.dispatchEvent(new Event(TranslateEvent)), []);
 
+  const swapLangText = () => {
```
sushain97, line 280: Let's use `React.useCallback()`.

src/components/translator/Translator.tsx
```diff
@@ -272,6 +277,11 @@ const Translator = ({ mode: initialMode }: { mode?: Mode }): React.ReactElement
...
+    setSrcText(tgtText);
+    setTgtText(srcText);
```
sushain97, line 282: Let's encapsulate the language switching here as well.

src/components/translator/Translator.tsx
sushain97: Let's also call this `swapLangs`. It could be used for document translation as well.

src/components/translator/Translator.tsx
sushain97: Moving the target text to the source makes sense. But, making the source text the target text doesn't make sense since Apertium's translators are not necessarily bijective. That is, just because "foo" in X translates into "bar" in Y doesn't mean "bar" in Y will translate into "foo" in X.

sushain97 (changes requested):

src/components/translator/Translator.tsx
satti-hari-krishna-reddy: Thanks for pointing that out. Yeah, it makes no sense to make the source text the target text

src/components/translator/Translator.tsx
satti-hari-krishna-reddy: I want to confirm if the suggestion is to define it globally, or if there's an alternative approach you recommend for maintaining it in a single place?

sushain97: You can define it one component (ideally the parent component) and export the constant for use in other files.

src/components/translator/Translator.tsx
```diff
@@ -272,6 +277,14 @@ const Translator = ({ mode: initialMode }: { mode?: Mode }): React.ReactElement
 
   const onTranslate = React.useCallback(() => window.dispatchEvent(new Event(TranslateEvent)), []);
 
+  const swapLangs = React.useCallback(() => {
+    setSrcText(tgtText);
+  }, [setSrcText, tgtText]);
+
+  const swapLangText = () => {
+    swapLangs();
+  };
```
sushain97, line 286: This function doesn't have any purpose now? I think we can just pass `swapLangs` around now. It's also a bit odd that `swapLangs` only has one thing happen inside it. Can we also move the actual language swap there instead of `src/components/translator/LanguageSelector.tsx`?

src/components/translator/Translator.tsx
```diff
@@ -272,6 +277,14 @@ const Translator = ({ mode: initialMode }: { mode?: Mode }): React.ReactElement
 
   const onTranslate = React.useCallback(() => window.dispatchEvent(new Event(TranslateEvent)), []);
 
+  const swapLangs = React.useCallback(() => {
+    setSrcText(tgtText);
```
sushain97, line 281: Does this line ever get invoked by tests?

sushain97 (changes requested):

satti-hari-krishna-reddy:
@sushain97
 i moved `swapLangs` from `src/components/translator/LanguageSelector.tsx` to the `src/components/translator/Translator.tsx` and The swapping of languages and text is functioning as expected.
However, this change has caused some issues in `LanguageSelector.test.tsx`. I made some adjustments to the test file to make it pass, but I think it might not be the optimal solution. Could you please take a look and help me on how to update the test accordingly?

src/components/translator/Translator.tsx
```diff
@@ -236,8 +244,18 @@ const WithTgtLang = ({
     },
     [pair, selectedPrefs, setSelectedPrefs],
   );
-
-  return children({ tgtLang, setTgtLang, recentTgtLangs, setRecentTgtLangs, pairPrefs, setPairPrefs });
+  const swapLangs = React.useMemo(
```
sushain97, line 247: This should be `React.useCallback`?

src/components/translator/__tests__/LanguageSelector.test.tsx
```diff
@@ -84,20 +85,19 @@ it('switches between mobile and desktop', () => {
 
 describe('swapping', () => {
   it('does not allow swapping when swapped pair invalid', () => {
-    renderLanguageSelector();
+    renderLanguageSelector({ swapLangs: undefined });
```
sushain97, line 88: I don't think this is testing what the test description describes anymore?

src/components/translator/__tests__/TextTranslationForm.test.tsx
```diff
@@ -20,13 +23,34 @@ const renderTextTranslationForm = (
...
+    setSrcText: jest.fn(),
+    setTgtText: jest.fn(),
     setLoading: jest.fn(),
     ...props_,
   };
 
+  const Wrapper = () => {
+    const [srcText, setSrcText] = useLocalStorage('srcText', '', {
+      overrideValue: getUrlParam(history.location.search, textUrlParam),
+    });
```
sushain97, line 37: Why do we need to duplicate this code in a test...?

src/components/translator/__tests__/TextTranslationForm.test.tsx
```diff
@@ -149,16 +173,31 @@ describe('translation', () => {
...
+      setTgtText: jest.fn(),
       setLoading: jest.fn(),
     };
 
     const Container = () => {
       const [srcLang, setSrcLang] = React.useState('eng');
+      const [srcText, setSrcText] = useLocalStorage('srcText', '', {
+        overrideValue: getUrlParam(history.location.search, textUrlParam),
+      });
+      const [tgtText, setTgtText] = React.useState('');
```
sushain97, line 188: Same question here as above.

sushain97 (changes requested):

src/components/translator/__tests__/TextTranslationForm.test.tsx
satti-hari-krishna-reddy: To be honest, I'm currently stuck here, and I'm not entirely sure why I went with this approach. It seems to be the only way the tests are passing, and I need your help on how to improve this setup. The functionality of `TextTranslationForm` is closely tied to state variables passed as props from its parent.

satti-hari-krishna-reddy:
@sushain97 
I'm relatively new to testing and haven't worked with it before.

src/components/translator/__tests__/TextTranslationForm.test.tsx
sushain97: The problem is that the tests need to be relocated now. Moving the logic that pulls `srcText` from local storage from `TextTranslationForm` to `Translator` means that the tests which verify that functionality need to be moved to `Translator.test.tsx`.

satti-hari-krishna-reddy: I realized that my knowledge of React testing might be a bit lacking. I plan to learn jest, and then I'll resume working on the pull request at a later point in time. I'll focus on other issues for now

PR #483, opened 2024-01-07
added a string directing to beta.apertium.org

satti-hari-krishna-reddy:
**Description:**
This PR addresses issue #361 and also builds upon and supersedes the changes proposed in PR #422.

**Changes Made:**
- Modified `config.ts` by adding `showMoreLanguagesLink` and specified its type as `boolean` in `types.ts`. Also, added `more_languages` to `stringReplacements`.
- Modified `AboutModal.tsx` by conditionally rendering `more_languages` based on the value of `shouldDisplayString`, either true or false.
- In `index.test.tsx`, modified a test to check if, when the About dialog is opened and `showMoreLanguagesLink` is set to `true`, it should render the `more_languages`. Similarly, when `showMoreLanguagesLink` is set to `false` and the About dialog is opened, it should not render the `more_languages`.

### DIFF
--- a/config.ts
+++ b/config.ts
@@ -14,5 +14,7 @@ export default {
 
   stringReplacements: {
     '{{maintainer}}': "<a href='https://wiki.apertium.org/wiki/Apertium' target='_blank' rel='noopener'>Apertium</a>",
+    '{{more_languages}}': "<a href='https://beta.apertium.org' rel='noreferrer'  target='_blank'>beta.apertium.org</a>",
   },
+  showMoreLanguagesLink: false,
 } as Config;
--- a/src/components/footer/AboutModal.tsx
+++ b/src/components/footer/AboutModal.tsx
@@ -3,6 +3,7 @@ import Modal, { ModalProps } from 'react-bootstrap/Modal';
 import Col from 'react-bootstrap/Col';
 import Row from 'react-bootstrap/Row';
 
+import { ConfigContext } from '../../context';
 import { useLocalization } from '../../util/localization';
 
 import alicanteLogo from './img/logouapp.gif';
@@ -18,6 +19,7 @@ import mineturLogo from './img/logomitc120.jpg';
 import prompsitLogo from './img/prompsit150x52.png';
 
 const AboutModal = (props: ModalProps): React.ReactElement => {
+  const config = React.useContext(ConfigContext);
   const { t } = useLocalization();
 
   return (
@@ -28,6 +30,7 @@ const AboutModal = (props: ModalProps): React.ReactElement => {
       <Modal.Body>
         <p dangerouslySetInnerHTML={{ __html: t('What_Is_Apertium') }} />
         <p dangerouslySetInnerHTML={{ __html: t('Maintainer') }} />
+        {config.showMoreLanguagesLink && <p dangerouslySetInnerHTML={{ __html: t('More_Languages') }} />}
 
         <Row className="lead">
           <Col md="6">
--- a/src/components/footer/__tests__/index.test.tsx
+++ b/src/components/footer/__tests__/index.test.tsx
@@ -2,6 +2,7 @@ import * as React from 'react';
 import { getAllByRole, render, screen, waitFor } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 
+import Config from '../../../../config';
 import Footer from '..';
 
 const renderFooter = () => {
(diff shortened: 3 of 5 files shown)


satti-hari-krishna-reddy:
hi @sushain97  , could you please review my pull request?  : )

config.ts
```diff
@@ -14,5 +14,8 @@ export default {
 
   stringReplacements: {
     '{{maintainer}}': "<a href='https://wiki.apertium.org/wiki/Apertium' target='_blank' rel='noopener'>Apertium</a>",
+    '{{more_languages}}': "<a href='https://beta.apertium.org' rel='noreferrer'  target='_blank'>beta.apertium.org</a>",
   },
+
+  shouldDisplayString: false,
```
sushain97, line 20: This should be more specific. Something like `showMoreLanguagesLink`?

src/components/footer/AboutModal.tsx
```diff
@@ -28,6 +29,9 @@ const AboutModal = (props: ModalProps): React.ReactElement => {
       <Modal.Body>
         <p dangerouslySetInnerHTML={{ __html: t('What_Is_Apertium') }} />
         <p dangerouslySetInnerHTML={{ __html: t('Maintainer') }} />
+        {Config.shouldDisplayString && (
+          <p dangerouslySetInnerHTML={{ __html: t('More_Languages') }} style={{ paddingBottom: '2em' }} />
```
sushain97, line 33: Rather than specific padding values, can you use Bootstrap's utility spacing classes?

src/components/footer/AboutModal.tsx
```diff
@@ -28,6 +29,9 @@ const AboutModal = (props: ModalProps): React.ReactElement => {
       <Modal.Body>
         <p dangerouslySetInnerHTML={{ __html: t('What_Is_Apertium') }} />
         <p dangerouslySetInnerHTML={{ __html: t('Maintainer') }} />
+        {Config.shouldDisplayString && (
```
sushain97, line 32: This should come from `React.useContext(ConfigContext)`.

src/strings/ara.json
```diff
@@ -51,6 +51,7 @@
     "Help_Improve": "ساعد(ي) في تحسين أبرتيُم!",
     "Contact_Us": "نرحب بتواصلك معنا في حالة العثور على خطأ، أو إن كان هناك مشروع تود(ين) أن تطرح(ي) علينا العمل عليه، أو إن أردتم المساعدة.",
     "Maintainer": "هذا الموقع يتكفل بصيانته {{maintainer}}.",
+    "More_Languages": "تبحث عن المزيد من اللغات؟ جرب {{more_languages}}",
```
sushain97, line 54: Do you know all of these languages? If not, you shouldn't be adding translations for all of them.

sushain97 (changes requested):

src/strings/ara.json
satti-hari-krishna-reddy: Ayo, my bad on that one. I didn't realize I shouldn't be adding translations for languages I'm not familiar with. Should've checked with you before making those changes. Thanks for pointing it out! I've reverted the changes as requested.

src/components/footer/__tests__/index.test.tsx
```diff
@@ -37,7 +38,19 @@ describe('Footer', () => {
   });
 
   describe('navigation buttons', () => {
-    it('opens about dialog', () => {
+    it('opens about dialog and display show more languages link when showMoreLanguagesLink is true', () => {
+      Config.showMoreLanguagesLink = true;
```
sushain97, line 42: Tests shouldn't be modifying the configuration directly. They should use `ConfigContext.Provider`. I think there's an example elsewhere?

src/strings/eng.json
```diff
@@ -48,6 +48,7 @@
     "Help_Improve": "Help us improve Apertium!",
     "Contact_Us": "Feel free to contact us if you find a mistake, there's a project you would like to see us work on, or you would like to help out.",
     "Maintainer": "This website is maintained by {{maintainer}}.",
+    "More_Languages": "Looking for more languages? Try {{more_languages}}",
```
sushain97, line 51: Should this include a `.` at the end? It'll be easier to tell if you include a screenshot in the PR description.

sushain97 (changes requested):
Thanks for backing out those changes! Almost there.

satti-hari-krishna-reddy:
I've implemented the changes  utilizing configContext. However, I'm uncertain if this is the correct approach. Please review and provide feedback when you have a moment.

satti-hari-krishna-reddy:
Screenshot (19)

sushain97 (approved):

Issue #362, opened 2020-02-23
Clicking on the Apertium logo should return user to the home page


woswos:
Conventionally clicking on the logo of a website takes users to the home page. When I click on the Apertium logo, nothing happens. I think this is a small but essential detail in the user interface. I will try to make this change and link PR to this issue. 

Is there a historical reason why the logo is not clickable?

sushain97:
Where would the logo take you? html-tools is a SPA.

woswos:
It should work like the current "cancel" button in Apertium's case. I forgot to mention it.

sushain97:
> It should work like the current "cancel" button in Apertium's case. I forgot to mention it.

🤷‍♂ Usually the logo-link is used as a way to provide a `Home` button. I'm not sure what benefit having both `Cancel` and the logo accomplish the same thing adds but I don't have a strong reason to oppose it either.

woswos:
I think it is really annoying and I still accidentally click the logo to go back. I think it is good to comply with the trends that improve user experience.

Also, IMO, calling the `Go Back` button `Cancel` is not right. AFAIK, `Cancel` buttons are _usually_ used for stoping users from possible accidental/harmful changes or actions they can do. 

Some examples [Situation] -> [Possible Buttons]: 
- "Do you want to delete file X?" -> `Cancel/No` `Yes/Confirm/Ok`
- "Do you want to proceed with the payment?" -> `Cancel/No` `Yes/Confirm/Ok`
- "Do you want to subscribe to the newsletter?" -> `Cancel/No` `Yes/Confirm/Ok`
- "_Operating system file open dialog_" -> `Cancel` `Open`
- "_Buttons at the end of a user profile_" -> `Cancel` `Update Profile`

<img width="1792" alt="Screen Shot 2020-02-24 at 12 39 00 AM" src="https://user-images.githubusercontent.com/10498874/75130786-1972d900-569e-11ea-82bd-1e33445a3900.png">

<img width="1792" alt="Screen Shot 2020-02-24 at 12 40 09 AM" src="https://user-images.githubusercontent.com/10498874/75130811-3ad3c500-569e-11ea-8e13-9e506f5431e0.png">

Even a quick Google or DuckDuckGo search shows red buttons with `X` symbol. There is no single `Cancel` button with `←` that connotes going back to the previous page/view. 

[...]

PR #491, opened 2024-03-14
fixed cancel button issue: changed "Cancel" to "Go Back"

gopalM-3:
fixes issue #362 by changing the 't' function's input text from "Cancel" to "Go Back" in both DocTranslationForm.tsx and WebpageTranslationForm.tsx


sushain97 (changes requested):
The `Go Back` string doesn't exist in any of the translation JSON files. In particular, it must exist in `eng.json`.

PR #494, opened 2024-03-17
Issue #488

### LINKED ISSUE #488: The copy option is on the  translated text
Screenshot 2024-02-23 211141
its a small one we can fix it

Cellzawy:
Attempted to fix #488 

Edited padding for `form-control` class in `bootstrap.css` to prevent the copy and clear buttons from being on top of the translated text. 

Little note: I am a little new to open source so excuse me if I did something wrong.


sushain97 (changes requested):
Please don't edit bootstrap.css. It's generated from upstream sources. There a number of other CSS files that can be edited instead with specific styles. Or, use inline utility classes.

Cellzawy:
> Please don't edit bootstrap.css. It's generated from upstream sources. There a number of other CSS files that can be edited instead with specific styles. Or, use inline utility classes.

Noted, sorry for that. Working on it now

Cellzawy:
I have removed my edits on bootstrap.css like you asked, added `pt-4` to both of the `Form.Control`s in `TextTranslationForm.tsx`
This should prevent the buttons from being on top of the text, let me know if you need more spacing though.

ahmedsiam0:
I think more spacing is needed.
When hovering, the button overlaps with the text.
hover

Cellzawy:
> I think more spacing is needed. When hovering, the button overlaps with the text. hover

Fair enough, to increase the spacing I would give it `pr-5` instead of `pr-4`. This is how it would look:
image

Or, if `pr-5` it's too much space I can create my own class (let's name it pr-4-5 for example) and give it a padding that's in between `pr-4` and `pr-5`. It would look like this:
image

Let me know which one you prefer.

ahmedsiam0:
The second option looks visually better for me. (between `pr-4` and `pr-5`).

Cellzawy:
> The second option looks visually better for me. (between `pr-4` and `pr-5`).

Done, waiting for your feedback.

src/app.css
```diff
@@ -12,3 +12,7 @@
 .btn-primary:hover {
   background-color: #385a7f;
 }
+
+.pr-4-5 {
+  padding-right: 2.25rem;
+}
```
sushain97, line 18: Let's add this to `translator.css` instead and give it a specific name rather than `pr-4-5`. Perhaps `.translation-text-input`.

sushain97 (changes requested):
I'm now seeing a large unnecessary padding on the translated text:

image

Cellzawy:
> I'm now seeing a large unnecessary padding on the translated text:
> 
> image

Right, then instead of using padding and absolute positioning, I am suggesting to make both the textarea and the button components sibling children of a parent div.

This is how it looks:
image
image

Let me know if you like this approach and I will push my changes.

Cellzawy:
> I'm now seeing a large unnecessary padding on the translated text:
> 
> image

Right, apparently that is because of the interface language which is RTL. I have missed that part, sorry.

Fixed it by setting the padding on the left if the interface language is RTL and the opposite if it's LTR:
image
image

Waiting for your feedback.

Cellzawy:
> I'm now seeing a large unnecessary padding on the translated text:
> 
> image

Right, apparently it's because of the difference in the interface language's direction, sorry I have missed that part.
Fixed it by setting padding on left if it's an RTL language and the opposite for LTR:
image
image

src/components/translator/translator.css
```diff
@@ -98,3 +98,11 @@ html[dir='rtl'] .translation-modes > :not(:first-child) {
...
   height: 700px;
 }
+
+html[dir='ltr'] .translation-text-input {
+  padding-right: 1.85rem;
+}
+
+html[dir='rtl'] .translation-text-input {
+  padding-left: 1.85rem;
+}
```
sushain97: Please add a trailing newline.

sushain97 (approved):

Issue #441, opened 2022-03-31
format long names in unconfusing way


jonorthwash:
Long language names end up looking like multiple language names.

![Screenshot_2022-03-31_19-22-28](https://user-images.githubusercontent.com/963849/161165364-96c913e9-290c-47b8-94ea-1c9eb00488df.png)

In this screenshot there are two of them.

I can think of two solutions:
* the non-first lines could be indented a bit
* some sort of dividers between/around the names?

unhammer:
Slightly unrelated, but wouldn't it be better to write it as `Nahuatl, Zacatlán-Ahuacatlán-Tepetzintla` so people scanning for Nahuatl find it amongst the N's?

jonorthwash:
Perhaps the same could be asked for Zapotec.  @ftyers, what do you think?

PR #497, opened 2024-03-21
Added dividers between languages

nrakhay:
Hello!

This is my solution to the **Issue #441** 
I tried both options proposed by Jonathan and,

I think we should go with adding a **divider between languages** as it looks cleaner and clearer. Below is the screenshot for this solution

Screenshot 2024-03-21 at 22 10 19

However, I think you should also take a look at how solution with divider looks like:

Screenshot 2024-03-21 at 22 13 17

### DIFF
--- a/src/components/translator/LanguageSelector.tsx
+++ b/src/components/translator/LanguageSelector.tsx
@@ -211,23 +211,28 @@ const LangsDropdown = ({
       const [code, name] = langs[j];
       const valid = !validLang || validLang(code);
       langElems.push(
-        <button
-          className={classNames('language-name', {
-            'variant-language-name': isVariant(code),
-            'text-muted': !valid,
-          })}
-          disabled={!valid}
-          key={code}
-          onClick={() => {
-            setLang(code);
-            document.body.click();
-          }}
-          tabIndex={0}
-        >
-          {name}
-        </button>,
+        <div>
+          <button
+            className={classNames('language-name', {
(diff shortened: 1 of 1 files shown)


nrakhay:
Also, I could not make it to creating an account inside the Appertium Wiki and MediaWiki (the latter says incorrect captcha although it is 100% correct). 

Is there any other way of creating it with the help of mentors?

sushain97 (changes requested):
This shouldn't require modifying the DOM at all. It should be possible with just CSS changes.

nrakhay:
I returned things how they were on the DOM level and added a divider using :after in CSS 

image

src/components/translator/LanguageSelector.tsx
```diff
@@ -228,6 +228,7 @@ const LangsDropdown = ({
         </button>,
       );
     }
+
```
sushain97, line 231: Why did this change?

sushain97 (commented):
This seems much more reasonable.

I think it should be a bit lighter.

I'd see what @jonorthwash thinks.

jonorthwash:
The dividers separating varieties (e.g., for Occitan, Panjabi, etc.) should be different from those separating languages.

I don't totally love separators, but I agree that something would be nice here.  It doesn't look as clean, even though it is clearer.  How does Google (or other bigger site) deal with this problem?

PR #501, opened 2024-07-10
display subreadings more sensibly on analyzer page

mr-martian:


mr-martian:
The example given in #8 now displays as

image

PR #515, opened 2025-08-25
i18n : Add Kabyle language

ghost:
Hi,

We would like to introduce kabyle `kab` language on Apertium HTML Tools.
It will be followed by some work on apertium-kab and some language pairs.

This translation was made by me and @ZiriSut

Related to : https://github.com/apertium/apertium-html-tools/issues/514

Regards,


unhammer:
@sushain97 that lint fail is not related, right?

sushain97:
Nope, it seems to have been introduced in https://github.com/apertium/apertium-html-tools/commit/33876a02c4de46c3f6caf33f198febae38a9296a by @TinoDidriksen. My guess is that the version of Typescript that was upgraded to in that change is incompatible with the version of react-bootstrap currently in use and a couple issues in the app itself.

unhammer (approved):

PR #519, opened 2025-10-15
Dictionary mode

vykliuk:


unhammer:
This is a huge set of changes, can you describe what it does? It changes both localisations and logic and presentation and build, are these changes all related to one feature?

config.ts
TinoDidriksen, line 1: The config can't be merged as-is. Don't change the default URL, mode, name, or color.

jonorthwash:
> This is a huge set of changes, can you describe what it does? It changes both localisations and logic and presentation and build, are these changes all related to one feature?

It's for Paradigm Dictionary mode:
https://wiki.apertium.org/wiki/Apertium-html-tools/Paradigm_dictionary

config.ts
jonorthwash: @vykliuk, yeah, revert your changes to the config file in this branch/PR, and just keep them locally.

jonorthwash: This was apparently entirely inadvertent.  I've reverted it at @vykliuk's request.

scripts/gen-lang-registry.ts
```diff
@@ -0,0 +1,29 @@
...
+      return `  ${JSON.stringify(name)}: ${name}Plugin,`;
+    })
+    .join('\n');
+
+  const content = `import Dictionary from './Dictionary';\n${imports}\nimport { LanguagePlugin } from './types';\n\nexport const languageRegistry: Record<string, LanguagePlugin> = {\n${registry}\n};\n\nexport default Dictionary;\n`;
+
+  fs.writeFileSync(INDEX_FILE, content);
+}
+
+generateRegistry();
```
unhammer, line 29: This seems to auto-create an index.ts which refers to all the files in langs, but if I read it correctly, the generated file will just contain something like 

```
import Dictionary from './Dictionary';
import { spaPlugin } from './langs/spa';
import { uumPlugin } from './langs/uum';
import { LanguagePlugin } from './types';

export const languageRegistry: Record<string, LanguagePlugin> = {
  "spa": spaPlugin,
  "uum": uumPlugin,
};

export default Dictionary;
```

I understand that this build step is here to avoid having to remember to do 
```diff
+import { tatPlugin } from './langs/tat';
…
+  "tat": tatPlugin,
```
on adding Tatar etc., but OTOH it adds some complexity to the build. Do we expect lots of `langs`?

(Does the regular localisation code also do this kind of thing?)

unhammer:
I haven't looked very deeply into the react code, @sushain97 knows that better, but I think from just reading it, it looks good to me, the new stuff seems pretty self-contained. I would prefer squashing into one commit for the Dictionary stuff and one for the fairly unrelated src/strings updates – or just not including the src/strings updates if they're just from running `src/strings/localisation-tools.py`.

scripts/gen-lang-registry.ts
jonorthwash: I think eventually we expect support for this to be available for lots of languages, yes.

On the other hand, it's not hard to list them all.  I did ask @vykliuk explicitly to add this feature (compiling the list automatically), as it makes everything language-specific contained in that one folder.  But I'm not super committed to keeping it—as you point out, it's not hard to add the language to this list.

